PR #4, opened 2015-10-04
fix const with an expression

rom1504:
Just added a minimal failing test to start with

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -160,7 +160,7 @@ module.exports = function(api, tokens, EOF) {
     /**
      * Reads a constant declaration
      * <ebnf>
-     *  constant_declaration ::= T_STRING '=' scalar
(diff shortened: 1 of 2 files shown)


rom1504:
https://github.com/glayzzle/php-parser/blob/1dc2b2e2946ec788de4d3fe78331d1424b1929c8/src/parser/class.js#L163 seems to be the pb

rom1504:
fixed

src/parser/class.js
```diff
@@ -169,7 +169,7 @@ module.exports = function(api, tokens, EOF) {
         .next()
         .expect('=')
         .next()
-        .read_scalar()
+        .read_expr()
```
ichiriac: https://github.com/php/php-src/blob/a49ce7bb91bec02d6f26b3118404371df23242fe/Zend/zend_language_parser.y#L801

PR #61, opened 2017-01-22
Improvements on AST nodes

ichiriac:
Thanks @motiz88, I appreciate your help


motiz88:
No problem @ichiriac, note that this is a breaking change though so should be semver-major.

ichiriac:
Right, I will release a 2.0.0-alpha version but the AST structure is not yet stable. 

I need to make a pass in order to be sure that's consistent. I've got other breaking changes to pass, the 2.0.0 version will be released when the global API will be stable.

PR #85, opened 2017-08-17
Filter out falsy nodes in blocks

### LINKED ISSUE #84: Null node found in short-form `if` body
In the following code (see also [AST Explorer snippet](https://astexplorer.net/#/gist/dcde248b9867b77c126e2e926dd06f43/4029a50b0d58062ef70097eb4208d89ab9680ca6)) the `if` statement's body is currently parsed as `{kind: "block", children: [null]}`.

```php
<?php if (true): ?>
<?php endif; ?>
```

I have a PR upcoming to fix and regression-test this.

motiz88:
Fixes #84 and adds a regression test. This is perhaps a crude fix, as I chose not to hunt down the specific `null`-producing cases in e.g. [`parser/if.js`](https://github.com/glayzzle/php-parser/blob/a99ae872cd33259d534c2631020482911b943368/src/parser/if.js), but rather to remove them after the fact in the `Block` constructor.

### DIFF
--- a/src/ast/block.js
+++ b/src/ast/block.js
@@ -15,7 +15,7 @@ var KIND = 'block';
  */
 var Block = Statement.extends(function Block(kind, children, location) {
   Statement.apply(this, [kind || KIND, location]);
-  this.children = children;
+  this.children = children.filter(Boolean);
 });
 
 module.exports = Block;
--- a/test/ifTests.js
+++ b/test/ifTests.js
@@ -102,4 +102,19 @@ describe('Test IF statements', function() {
       ast.children[1].alternate.kind.should.be.exactly('block');
     });
   });
+  describe('Issue #84', function () {
+    var ast = parser.parseCode([
+      '<?php if (true): ?>',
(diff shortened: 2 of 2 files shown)


ichiriac:
Hi @motiz88,

Thanks for the fix. I've digged into the issue, it's caused by this code : https://github.com/glayzzle/php-parser/blob/master/src/parser/statement.js#L317

It can also be reproduced by this code :

```php
<?php if (true): ; ; ?>
<?php endif; ?>
```

The actual fix works on every case, but may slow down a bit the AST building process. I'm looking at a more early filtering solution.

On the other hand, the bracket `{ ... }` form does not have the problem because the falsy values are filtered here : https://github.com/glayzzle/php-parser/blob/master/src/parser/statement.js#L83 

I've also found a helper here https://github.com/glayzzle/php-parser/blob/master/src/parser/utils.js#L15 used on loops (could be fixed and used with multiple tokens)

Do you want to digg more ? 

If you don't have the time I can merge the fix and continue to correct accordingly.

PR #92, opened 2017-10-08
Added reserved word support for method and constant names

bbtgp:
This covers all reserved word as names in class/trait methods and constants.


ichiriac:
Seriously, I'm amazed :+1: - everything is OK. Now I need to make them pass the [php-parser-dev](https://github.com/glayzzle/php-parser-dev} test suite

ichiriac:
[build fails](https://travis-ci.org/glayzzle/php-parser-dev/builds/284992680) on `foo::echo` but it's normal, I will update this.

Got also others warnings but they were here before and may be real parse errors from the frameworks. Do you know other frameworks that uses a lot of PHP7 language features ?

I need to read more the PHP7 yaccs in order to find edge cases but everything passes without any regression so we can release it under a stable version as `2.1.0`  :+1:

PR #93, opened 2017-10-08
Added support for static lookup on property

bbtgp:
Example
```php
$this->foo::bar;
```

I don't see any rfc or documentation for this feature but it does work in php 7.0, iv been using it in production for quit a while.


ichiriac:
Hi @bbtgp, thanks good job !

They changed the name of the token various times between `T_DOUBLE_COLON` and `T_PAAMAYIM_NEKUDOTAYIM`, here a discussion about it : https://philsturgeon.uk/php/2013/09/09/t-paamayim-nekudotayim-v-sanity/

So here a list of classes used before T_DOUBLE_COLON :

- class_name 
- variable_class_name 
- new_variable ([simple_variable T_OBJECT_OPERATOR property_name](https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1170))

So you are right, the syntax is valid until it's prefixed by [simple_variable](https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1150) or a |class_name](https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1036)

Also the variable_class_name (alias [dereferencable](https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1115)), allows to use it after a function call :

```
echo $this->foo()::bar;
echo foo()::bar;
echo foo::bar[
```

But in order to check if the algorithm is not too permissive, should also fail parsing :

```
echo foo()::bar->foo;
echo foo::bar->foo;
```

Do you want that I merge it now, or do you need changes ?

bbtgp:
Ok, this should deal with dynamic functions and array offset lookup in constants. Property lookup on constant and static lookup on constant should fail appropriately.

Coverage decreased so I might look at that later, but ill be happy to put more time into it next weekend.

ichiriac:
Hi @bbtgp, I've tested and everthing works. As it adds missing support for PHP 7, but keeps backward compatibility, I will release it under a 2.1 version.

Many thanks for your contribution :heart:

bbtgp:
Ok cool. I just wanted to mention that this supports lookup of the ::class constant which it should not. I should of pushed it last week, sorry about that!

ichiriac:
Not sure to understand, the `fooBar::class` lookup is supported in PHP since version 5.5. Maybe you refer to constants declaration like this :

```php
<?php
class Foo {
  const class = 'Foo'; // Fatal error
}
```

This kind of error should be raised at parsing time. Also got another thing to port on methods names :

```c
if(ZEND_NOT_RESERVED != zend_check_reserved_method_name(decl->name)) {
  zend_error_noreturn(E_COMPILE_ERROR,
    "Cannot use '%s' as class method name as it is reserved", decl->name->val);
}
```

Also saw a tiny improvement here : https://github.com/glayzzle/php-parser/blob/master/src/parser/class.js#L186

Instead `if (this.is('IDENTIFIER') || this.expect(this.tok.T_STRING))` it should be faster to do `if (this.token === this.tok.T_STRING || this.expect('IDENTIFIER'))` because T_STRING verification takes way less time to operate and occurs more often ...

If you want to I can wait a new PR, but if don't have time tell me and I will make a bugfix release soon on these.

bbtgp:
My bad then, i was referring to 'Dynamic class names are not allowed in compile-time' when running something like $test->test::class;

I can make that change this weekend maybe, I have no free time during the week though

PR #94, opened 2017-11-04
Added suggested performance optimization

bbtgp:
I switched the expressions around in the conditions for class.js and variable.js and removed the T_CLASS check that was not needed.

Cheers!

### DIFF
--- a/src/parser/variable.js
+++ b/src/parser/variable.js
@@ -127,23 +127,20 @@ module.exports = {
           result = node(result, offset);
           break;
         case this.tok.T_DOUBLE_COLON:
-          var node = this.node('staticlookup'), offset;
-          this.next();
-
-          if(this.is('IDENTIFIER') || this.token === this.tok.T_STRING
-              || this.token === this.tok.T_CLASS
-          ) {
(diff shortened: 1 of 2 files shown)


src/parser/variable.js
```diff
@@ -127,23 +127,20 @@ module.exports = {
...
+            var offset = this.node('constref');
             var name = this.text();
+            
             this.next();
             offset = offset(name);
-
+            
             if(this.token === this.tok.T_OBJECT_OPERATOR || this.token === this.tok.T_DOUBLE_COLON) {
               this.error();
+            } else {
```
ichiriac: possible infinite loop and node may missing in graceful mode

ichiriac:
Hi @bbtgp, 

Many thanks :+1:  

There will be just a little problem on graceful mode (not sure, not tested yet).  I don't want to abuse your time so tell me if you're done, I can patch it myself if needed.

ichiriac:
Should pass now, I just need to add some regression tests on it, I'll merge it soon and release it under 2.1.1 as a bugfix

Issue #50, opened 2017-01-02
Cleanup the code


ichiriac:
- [x] Add "use strict" statement on every file
- [x] Check & investigate each @todo / @fixme statement
- [x] Pass with JSHint on every file
- [x] Use ES6
- [x] Add a transpiler for ES5 support

ichiriac:
Hi @czosel,

I would be really glad if you can help on this. When the project was started es6 was not mature but I agree that I should start to migrate on it.

I have no problem with ES6 (to be honest I do not like the class support as it shadows the prototypal power of JS), but I would be glad to use all ES6 goodies. 

I just need to support browser with babel for example, and be compliant with node v4.

czosel:
Hi @ichiriac, 
Awesome! I'll definitely try to make some first steps in this direction. I actually also don't like ES6 classes too much myself, but it's hard to live without many of the other features (say arrow functions). Since you have a build step already, compiling down to ES5 for support in IE11/Node v4 should be no problem. Rewriting the whole thing to ES6 is probably a big task, so I'd suggest introducing some ESLint rules that throw warnings when old syntax is found. What do you think?

ichiriac:
Agree with you, as sometimes big refactoring can lead to regressions. The ESLint automated check is a must have. You can start by providing just a simple configuration that checks common rules and attach it on travis build steps :smile:

PR #102, opened 2017-12-26
Use short form for type casts

czosel:
The `type` of types casts has been inconsistent: Sometimes, an alias/long form was used (e.g. `boolean`), other times the short form (e.g. `int`). This PR changes this by always using the short form.
Please also see https://github.com/prettier/prettier-php/pull/5#discussion_r158664155 for some more context. 
Thanks for your feedback in advance!

### DIFF
--- a/src/parser/expr.js
+++ b/src/parser/expr.js
@@ -272,7 +272,7 @@ module.exports = {
         return this.node('cast')('int', this.next().read_expr());
 
       case this.tok.T_DOUBLE_CAST:
-        return this.node('cast')('double', this.next().read_expr());
+        return this.node('cast')('float', this.next().read_expr());
 
       case this.tok.T_STRING_CAST:
         return this.node('cast')('string', this.next().read_expr());
@@ -284,7 +284,7 @@ module.exports = {
         return this.node('cast')('object', this.next().read_expr());
 
       case this.tok.T_BOOL_CAST:
-        return this.node('cast')('boolean', this.next().read_expr());
+        return this.node('cast')('bool', this.next().read_expr());
 
       case this.tok.T_UNSET_CAST:
         return this.node('unset')(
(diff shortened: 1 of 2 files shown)


czosel:
Hi @ichiriac, a little question on the side: I noticed you're not using ES6 features - is that intentional? I also saw #50 - maybe it would be a good idea to introduce eslint, prettier, and slowly move towards ES6?

src/parser/expr.js
```diff
@@ -272,7 +272,7 @@ module.exports = {
         return this.node('cast')('int', this.next().read_expr());
 
       case this.tok.T_DOUBLE_CAST:
-        return this.node('cast')('double', this.next().read_expr());
+        return this.node('cast')('float', this.next().read_expr());
```
ichiriac: not sure in PHP but in principle double and float have not the same precision, I'll check this

test/exprTests.js
```diff
@@ -142,28 +142,35 @@ describe('Test expressions', function() {
   });
 
   it('test cast', function() {
+    var typesWithAlias = {
```
ichiriac: :+1:

src/parser/expr.js
ichiriac: :smile: got it from https://github.com/glayzzle/php-parser/blob/master/src/lexer.js#L111, actually there is only one token to these 3 casts types

czosel: Exactly - they are just aliases in PHP :wink:

ichiriac:
hi @czosel, I'll release it under 2.2.0 soon, I need to fix another little regression elsewhere, almost done

czosel:
Awesome! Thanks to your support I'm positive we'll have a working prettier implementation for PHP soon!

PR #109, opened 2017-12-28
Use let or const instead of var

czosel:


src/parser/expr.js
```diff
@@ -204,28 +204,28 @@ module.exports = {
       case this.tok.T_NEW:
         return this.next().read_new_expr();
 
-      case this.tok.T_ISSET:
+      case this.tok.T_ISSET: {
```
ichiriac: why surrounding with `{` ... `}` ? It it a linter directive or is it better ?

src/ast/variadic.js
```diff
@@ -4,19 +4,19 @@
...
 /**
  * Introduce a list of items into the arguments of the call
- * @constructor Variadic
+ * @constructor variadic
  * @extends {Expression}
  * @property {Array|Expression} what
  * @see https://wiki.php.net/rfc/argument_unpacking
  */
-var Variadic = Expr.extends(function Variadic(what, location) {
+const variadic = Expr.extends(function variadic(what, location) {
```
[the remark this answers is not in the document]
czosel: of course, this was a search/replace mistake :wink:

src/parser/expr.js
czosel: Yeah, eslint complains when you put `let`/`const` inside `case`. see https://eslint.org/docs/rules/no-case-declarations

czosel:
:smile: Thanks for the invite!

ichiriac:
You are very welcome my friend, I'm glad for your help :+1:

PR #112, opened 2018-01-26
Indent support for comments extraction

MyroshnykRoman:
Hi,
thank you for a great module.
I am using it for my project and faced a problem that ast extracted comments lose their indent, so after recompilation docs become hard to read.
I added support for indent.

Would be kind of you if you check it and approve.
extractDocWithIndent is a flag in options.

Best regards, Myroshnyk Roman

parser: {
      extractDoc: true,
      suppressErrors: true,
      php7: true,
      extractDocWithIndent:false
    },
    ast: {
      withPositions: false
    }

### DIFF
--- a/src/parser/comment.js
+++ b/src/parser/comment.js
@@ -6,6 +6,8 @@
 
 const docSplit = /^(\s*\*[ \t]*|[ \t]*)(.*)$/gm;
 
+const docSplitLeaveSpaces = /^(\s*[\*])(.*)$/gm;
+
 module.exports = {
   /**
    *  Comments with // or # or / * ... * /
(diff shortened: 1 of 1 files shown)


ichiriac:
Hi @MyroshnykRoman,

Thanks for the pull request. In order to merge it I need that the code behavior is tested on unit tests. Could you please provide a test in order to test the option and the AST output.

I've do not documented it yet, but with options you can also extend the parser functions, and make it behave like you want without needing to make a pull request.

 Take a look here : https://github.com/glayzzle/php-transpiler/blob/master/src/index.js#L89

I've injected new tokens, new ast nodes, and updated the rules of the parser. You can make almost what you want by extending it like this.

ichiriac:
Hi @MyroshnykRoman,

It will be corrected in the future 3.0 release by default without adding a new option -> avoid the trim command by default, so I close your PR.

PR #142, opened 2018-03-24
Add support for array destructuring

### LINKED ISSUE #121: Support array destructuring
```php
[$id1, $name1] = $data[0];
```

```php
["id" => $id1, "name" => $name1] = $data[0];
```

Ref: http://php.net/manual/en/migration71.new-features.php

czosel:
I finally dared (and found the time) to take a closer look at the parser :smile: I'm curious to hear your feedback!

Fixes #121

### DIFF
--- a/src/parser/expr.js
+++ b/src/parser/expr.js
@@ -185,6 +185,31 @@ module.exports = {
       }
     }
 
+    if (this.token === "[") {
+      expr = this.read_array();
+      if (this.token === "=") {
+        return this.node("assign")(expr, this.next().read_expr(), "=");
+      } else {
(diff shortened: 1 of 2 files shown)


ichiriac:
Hi @czosel,

I've read carefully your changes, and the algorithm is perfect :100: 

I only need to make a small change, I need that array deferencing behaves as `list`, so I'll use the same node in order to make it more easier to handle nodes translations for example (and keep a property for the semantics part)

Again, many thanks for your help

czosel:
☺️ great! Thanks for the review and the merge - I’ll take a look at your changes concerning the list node.

PR #143, opened 2018-03-24
Enhance AST representation of curlies / encapsed strings

### LINKED ISSUE #101: Curly braces in encapsed string, again
When working on prettier support for PHP (#100) i noticed something odd about curly braces in encapsed strings:
```php
"Hello {$foo}";
```
yields
```javascript
[ String {
            kind: 'string',
            loc: 
             Location {
               source: null,
               start: Position { line: 1, column: 1, offset: 1 },
               end: Position { line: 1, column: 7, offset: 7 } },
            value: 'Hello ',
            isDoubleQuote: false },
          Variable {
            kind: 'variable',
            loc: 
             Location {
               source: null,
               start: Position { line: 1, column: 8, offset: 8 },
               end: Position { line: 1, column: 12, offset: 12 } },
            name: 'foo',
            byref: false,
            curly: false } ],
```
But 
```php
"Hello ${foo}";
```
yields 
```javascript
[ String {
            kind: 'string',
            loc: 
             Location {
               source: null,
               start: Position { line: 1, column: 1, offset: 1 },
               end: Position { line: 1, column: 7, offset: 7 } },
            value: 'Hello ',
            isDoubleQuote: false },
          Variable {
            kind: 'variable',
            loc: 
             Location {
               source: null,
               start: Position { line: 1, column: 7, offset: 7 },
               end: Position { line: 1, column: 13, offset: 13 } },
            name: 'foo',
            byref: false,
            curly: true } ],
```
In the first case, the information about the surrounding curly braces seems to be lost, which makes it hard to re-print the node the way it was found. 
What do you think about this? If there is no semantic difference between the two different styles, we might get away with setting `curly:true` in both cases?

czosel:
Fixes #101

### DIFF
--- a/src/parser/scalar.js
+++ b/src/parser/scalar.js
@@ -199,7 +199,7 @@ module.exports = {
           this.expect("]") && this.next();
           name = node(name, offset);
         } else {
-          name = varName;
+          name = this.node("constref")(varName);
         }
       } else {
         name = this.read_expr();
@@ -210,6 +210,9 @@ module.exports = {
       // expression
       // https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1246
       result = this.next().read_variable(false, false, false);
+      if (result.kind === "variable") {
+        result.curly = true;
+      }
       this.expect("}") && this.next();
     } else if (this.token === this.tok.T_VARIABLE) {
       // plain variable
--- a/test/stringTests.js
+++ b/test/stringTests.js
@@ -1,28 +1,41 @@
 var parser = require("../src/index");
 
 describe("Test strings", function() {
-  it("fix #124", function() {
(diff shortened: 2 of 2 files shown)


src/parser/scalar.js
```diff
@@ -188,18 +188,17 @@ module.exports = {
       name = null;
       if (this.next().token === this.tok.T_STRING_VARNAME) {
         const varName = this.text();
-        name = this.node("variable");
```
ichiriac: I need to start the variable node builder before the next statement in order to correctly locate its offset

czosel: Oh, that explains why it was written that way! I reverted the change.

czosel:
@ichiriac let me know in case anything else needs to be changed :smile:

ichiriac:
It's all good, merged :+1:

PR #211, opened 2018-10-25
fix: location

### LINKED ISSUE #204: [bug] invalid start offset encapsed
Input:
```php
call(`
command;
command;
command;
`);
```

Has `13` offset, but should `12`.

For `string` it is `12`.

```php
call('
command;
command;
command;
');
```

evilebottnawi:
Not all, but better.

fixes #204

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -68,7 +68,7 @@ module.exports = {
 
       // check T_USE trait
       if (this.token === this.tok.T_USE) {
-        result = result.concat(this.next().read_trait_use_statement());
+        result = result.concat(this.read_trait_use_statement());
         continue;
       }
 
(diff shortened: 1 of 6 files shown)


evilebottnawi:
Hm, it is require to fix all problem with `codeclimate` right now?

src/parser/class.js
```diff
@@ -68,7 +68,7 @@ module.exports = {
 
       // check T_USE trait
       if (this.token === this.tok.T_USE) {
-        result = result.concat(this.next().read_trait_use_statement());
+        result = result.concat(this.read_trait_use_statement());
```
ichiriac: perfect

src/parser/class.js
```diff
@@ -365,6 +365,7 @@ module.exports = {
   read_trait_use_statement: function() {
     // defines use statements
     const node = this.node("traituse");
+    this.expect(this.tok.T_USE) && this.next();
```
ichiriac: top

ichiriac:
no need to fix the code climate, the most important part is the code coverage.

Issue #185, opened 2018-09-19
[bug] parens and staticlookup


evilebottnawi:
```php
$var = ($var[0])::foo;
$var = ($var[0][1])::foo;
$var = ($var[0])[1]::foo;
$var = (($var[0])[1])::foo;
$var = (new Foo())::bar;
```

Problem exists in master

evilebottnawi:
Here other big bug 
```php
get_class($this->resource)::$wrap;
```

ichiriac:
not sure yet, but seems that this bug existed before

evilebottnawi:
maybe :disappointed:

evilebottnawi:
Need also fix:
```php
get_class($var)::$$$$$property;
```

ichiriac:
Ok, for $$$$$property I'm not enough permissive here : https://github.com/glayzzle/php-parser/blob/master/src/parser/variable.js#L154

ichiriac:
other bugs are from : https://github.com/glayzzle/php-parser/blob/master/src/parser/variable.js#L232 and introduced here https://github.com/glayzzle/php-parser/issues/179 

I've forced the call as the bugfix request was over staticlookup with call functions, but here we are working on another branch :

```
simple_variable:
		T_VARIABLE			{ $$ = $1; }
	|	'$' '{' expr '}'	{ $$ = $3; }
	|	'$' simple_variable	{ $$ = zend_ast_create(ZEND_AST_VAR, $2); }
;
static_member:
		class_name T_PAAMAYIM_NEKUDOTAYIM simple_variable
			{ $$ = zend_ast_create(ZEND_AST_STATIC_PROP, $1, $3); }
	|	variable_class_name T_PAAMAYIM_NEKUDOTAYIM simple_variable
			{ $$ = zend_ast_create(ZEND_AST_STATIC_PROP, $1, $3); }
;
```

In principle it's fixed, but can you please provide more tests and check if it's not too permissive.

ichiriac:
released in `3.0.0-prerelease.8`

PR #214, opened 2018-10-30
[WIP] fix: respect `call` inside `staticlookup`

evilebottnawi:
partial https://github.com/glayzzle/php-parser/issues/185

### DIFF
--- a/src/parser/variable.js
+++ b/src/parser/variable.js
@@ -100,6 +100,75 @@ module.exports = {
     return result(what, offset);
   },
 
+  read_what: function(is_static_lookup = false) {
+    let what = null;
+    let name = null;
+    switch (this.next().token) {
+      case this.tok.T_STRING:
+        what = this.node("identifier");
+        name = this.text();
+        this.next();
+        what = what(name);
+
+        if (is_static_lookup && this.token === this.tok.T_OBJECT_OPERATOR) {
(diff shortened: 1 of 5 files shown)


src/parser/variable.js
```diff
@@ -100,6 +100,74 @@ module.exports = {
     return result(what, offset);
   },
 
+  read_what: function(name, is_static_lookup = false) {
```
ichiriac: the name variable is never used

src/parser/variable.js
```diff
@@ -140,98 +208,24 @@ module.exports = {
...
-              this.error([this.tok.T_STRING, this.tok.T_VARIABLE]);
-              // graceful mode : set what as error mode & continue
-              what = this.node("identifier");
-              name = this.text();
-              this.next();
-              what = what(name);
-              break;
-          }
-          result = node(result, what);
+          result = node(result, this.read_what(name));
```
ichiriac: the name paramter is extra, not used

ichiriac:
ok, the read_what function is correct, but you may only avoid name as parameter as it's not used

evilebottnawi:
@ichiriac WIP

evilebottnawi:
done :star:

ichiriac:
its all right

Issue #181, opened 2018-09-12
strange with parens and unary


evilebottnawi:
Input:
```php
$a = (-100);
$a = -(100);
```
First is `number` with `-`, second is `unary` operation.

How this do `babylon` https://astexplorer.net/#/gist/09d3bbd44f6274b78370283e4b780e68/c641a63b7a2a9ed776568165db4d2af13341738e

ichiriac:
:+1: I'll upgrade it

evilebottnawi:
Looks we fix it and released :+1: So we can close issue.

PR #215, opened 2018-10-30
fix: use unary instead negative number

evilebottnawi:
fixes #214

After investigate `babylon` and `PHP` parser (https://github.com/nikic/PHP-Parser) - they do not have negative number, they interpreted `-` as `unary` always, i think we should use same logic.

More tests included


ichiriac:
related / fixes : #181

PR #216, opened 2018-11-02
refactor: ast

evilebottnawi:
Remove `sys` because it is invalid:
- `empty` is not `Statement`, it is `Expression`
- `list` is not `Statement`, it is `Expression`
- `print` is not `Statement`, it is `Expression`
- `isset` is not `Statement`, it is `Expression`
- `empty` can't contain multiple arguments
- rename `args` in list to `items`, because it is items (https://github.com/nikic/PHP-Parser also use `items`)
- And other ast fixes, better look in code

More tests


src/ast/method.js
```diff
@@ -5,7 +5,7 @@
  */
 "use strict";
 
-const fn = require("./function");
+const Function = require("./function");
```
ichiriac: Function is a global declaration - in some cases (don't remember well where) this kind of declaration can cause errors

evilebottnawi: @ichiriac yes, you are right :+1:

PR #221, opened 2018-11-06
fix: reference node

evilebottnawi:
Includes:
- Add `raw` property (`self`, `parent`, `static` references can be in uppercase, mixedcase - `pArEnt`)
- `callable` and `iterable` to `typereference`
- fix detection `self` and `parent` ( as i write above `self` and `parent` can be in any case)
- Tests


ichiriac:
perfect :+1:

PR #223, opened 2018-11-06
fix: broken code

evilebottnawi:
Now we have broken code in `master` if parent script use `"use_strict";` - https://github.com/glayzzle/php-parser/blob/master/src/parser/statement.js#L253 (here `result` is not null, it is https://github.com/glayzzle/php-parser/blob/master/src/parser/statement.js#L339) to avoid these problem in future i will setup `eslint` and refactor some places

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,7 @@ module.exports = {
     sourceType: "module"
   },
   plugins: ["prettier"],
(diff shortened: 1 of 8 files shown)


evilebottnawi:
Can't update `prettier` plugin, need merge asap :disappointed:

ichiriac:
ok, got it, variables collisions

PR #227, opened 2018-11-06
fix: ast for constants

### LINKED ISSUE #210: [bug] invalid ast for `const`
Input:
```php
const CONSTANT = 'Hello', TEST = 'world';
```

Output each `const` as separate item, but should have same ast as for `static`. I will look how we can fix it.

evilebottnawi:
fixes #210

- Introduce `ConstantStatement` node
- `constant` node now have only `name` and `value`
- `ClassConstant` extends `ConstantStatement` and parse flags

### DIFF
--- a/src/ast/constantstatement.js
+++ b/src/ast/constantstatement.js
@@ -1,0 +1,25 @@
+/**
+ * Copyright (C) 2018 Glayzzle (BSD3 License)
+ * @authors https://github.com/glayzzle/php-parser/graphs/contributors
+ * @url http://glayzzle.com
+ */
+"use strict";
+
+const Statement = require("./statement");
+const KIND = "constantstatement";
+
+/**
+ * Declares a constants into the current scope
+ * @constructor ConstantStatement
+ * @extends {Statement}
+ * @property {Constant[]} items
+ */
+module.exports = Statement.extends(KIND, function ConstantStatement(
+  kind,
+  items,
+  docs,
+  location
+) {
+  Statement.apply(this, [kind || KIND, docs, location]);
+  this.items = items;
+});
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -51,8 +51,12 @@ module.exports = {
         return this.read_trait();
       case this.tok.T_USE:
         return this.read_use_statement();
-      case this.tok.T_CONST:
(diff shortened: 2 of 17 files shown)


src/ast/constantstatement.js
```diff
@@ -0,0 +1,25 @@
...
+const KIND = "constantstatement";
+
+/**
+ * Declares a static variable into the current scope
+ * @constructor Static
+ * @extends {Statement}
+ * @property {Variable[]|Assign[]} items
+ */
+module.exports = Statement.extends(KIND, function ConstantStatement(
+  kind,
```
ichiriac: here it means you either pass the kind from a parent class (inheritance), or null it you want to use current node kind (constantstatement).

src/parser/statement.js
```diff
@@ -51,8 +51,12 @@ module.exports = {
         return this.read_trait();
       case this.tok.T_USE:
         return this.read_use_statement();
-      case this.tok.T_CONST:
-        return this.next().read_const_list();
+      case this.tok.T_CONST: {
+        const result = this.node("constantstatement");
+        const items = this.next().read_const_list();
+        this.expectEndOfStatement();
+        return result(items);
```
ichiriac: here you need on the first parameter to pass null (related to the kind parameter)

src/ast/constantstatement.js
evilebottnawi: @ichiriac already find problem, thanks :+1:

src/parser/statement.js
evilebottnawi: @ichiriac fixed already :+1:

src/parser/statement.js
```diff
@@ -51,8 +51,13 @@ module.exports = {
         return this.read_trait();
       case this.tok.T_USE:
         return this.read_use_statement();
-      case this.tok.T_CONST:
-        return this.next().read_const_list();
+      case this.tok.T_CONST: {
+        const result = this.node("constantstatement");
+        const items = this.next().read_const_list();
+        this.expectEndOfStatement();
+        console.log(result);
```
ichiriac: oups, bad commit ? :smile:

evilebottnawi: @ichiriac yep, WIP on this :+1:

PR #232, opened 2018-11-10
fix: avoid `null` on last element in `array`/`list`

### LINKED ISSUE #174: Don't output `null` for last element in `list/array`
Input:
```
list(,$first, $second, $three, $fourth,) = $arr;
```

Contain `null` in `arguments`.

Babylon example: https://astexplorer.net/#/gist/6d37953c6b42f02a3ffe48b6101127c3/745b62f49d2c011d789e6d9754a4d705db5cdab0

I think it is unnecessary, because it is doesn't have sense for executing/parsing/printing/etc.

evilebottnawi:
fixes #174

### DIFF
--- a/test/snapshot/list.test.js
+++ b/test/snapshot/list.test.js
@@ -1,4 +1,4 @@
-const parser = require('../main');
+const parser = require("../main");
 
 describe("Test list expressions", function() {
   it("test list statements", function() {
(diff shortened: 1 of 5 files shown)


test/snapshot/list.test.js
```diff
@@ -31,4 +36,26 @@ describe("Test list expressions", function() {
...
+  it("fix #174 (2)", function() {
+    expect(
+      parser.parseEval(
+        "list($foo, $bar, $baz, list(, $nestedFoo,,, $nestedBar,)) = $array;"
+      )
+    ).toMatchSnapshot();
+  });
+
+  it("fix #174", function() {
+    expect(parser.parseEval("list($var,,,,,) = $array;")).toMatchSnapshot();
```
ichiriac: in that case it's stripping only the last one, so each time prettier will pass on the file it will remove one last null until it stips every null - sounds weird

evilebottnawi: @ichiriac hm, maybe you are right, need tests

evilebottnawi:
@ichiriac sounds reasonable, i look what i can do here

ichiriac:
Hi @evilebottnawi,

This PR starts to have conflicts, what about to close it and open an issue instead, then we will implement it later.

evilebottnawi:
Yep, let's do it, issue here https://github.com/glayzzle/php-parser/issues/174

PR #242, opened 2018-11-12
feat: improve `Declare` node and introduce `DeclareDirective` node

evilebottnawi:
- Introduce `DeclareDirective` node ( `key` (is `identifier`)/`value`)
- Rename `what` to `directives` for `Declare` node
- `Declare` can have multiple directive (now it is array).
- Tests


ichiriac:
Maybe the `Entry` node from array could also inherit from node same as the `DeclareDirective`

PR #249, opened 2019-01-18
fix: location for staticlookup and offsetlookup

evilebottnawi:
We already do same for `propertylookup`. It is allow for prettier format comment like:

```php
$var = $var
    // Comment
    ::each() // Comment
    // Comment
    ::map() // Comment
    // Comment
    ::first() // Comment
    // Comment
    ::dump();
```

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -369,7 +369,7 @@ AST.prototype.prepare = function(kind, docs, parser) {
     }
     AST.stack[++AST.stackUid] = {
       position: start,
-      stack: (new Error()).stack.split("\n").slice(3, 5)
+      stack: new Error().stack.split("\n").slice(3, 5)
     };
     result.stackUid = AST.stackUid;
   }
@@ -410,7 +410,7 @@ AST.prototype.prepare = function(kind, docs, parser) {
         parser._docIndex = parser._docs.length - docs.length;
       }
     }
-    if(parser.debug) {
+    if (parser.debug) {
       delete AST.stack[result.stackUid];
     }
   };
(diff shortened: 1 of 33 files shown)


src/ast.js
```diff
@@ -176,10 +176,14 @@ AST.prototype.swapLocations = function(target, first, last, parser) {
...
     target.loc.end = last.loc.end;
     if (this.withSource) {
-      target.loc.source = parser.lexer._input.substring(
-        target.loc.start.offset,
-        target.loc.end.offset
-      );
+      const startOffset = target.loc.start.offset;
+      let endOffset = target.loc.end.offset;
+
+      if (target.kind === "offsetlookup") {
```
ichiriac: could it be fixed before in order to keep it more simple ? I'll take a look where to fix it - like this here it will be more generic

evilebottnawi: @ichiriac hm, i will look, but it is strange, this code fix problem with `]`, without this code we remove last `]` from `offsetlookup`

evilebottnawi: @ichiriac don't think it can be fixed in other places, because we don't include `[` and `]` in `offset` for `offsetlookup` node (and it is expected, other parsers like babelon and etc use same logic), so we need add `+1` here to avoid location problem for `offsetlookup` node

evilebottnawi: @ichiriac hm, yep, i think how we can fix it

evilebottnawi: Don't have idea how we can fix it simple, we don't have information about where character `[` is starting and `]` is ending :disappointed: 

Simple example:
```php
$var = $var[ 'qqq' ]  ;
```

Modify location for `offset` property is not right solution.

ichiriac: I'm back tomorrow and I'll try to fix it

ichiriac: Hi @evilebottnawi - the swapLocations was not the right solution, I've created a different version resolveLocations.

Here the explanation :

## with resolveLocations :
- You have target => `['foo']`
- You have first => `$var`
- You have last => `'foo'`

The system will include first + last + target in order to cover all the source code, resulting in `$var['foo']`

## with swapLocations over `1 + 2 * 3` :

- You have target => `1 + 2`
- You have first => `2`
- You have last => `3`

It's used after switching nodes, so only first & last are resolved : `2 * 3`

PR #257, opened 2019-01-25
fix: use PropertyStatement node for multiple properties

### LINKED ISSUE #256: invalid ast for multiple properties
Input:
```php
class Foo {
    private $dsn, $username, $password;
}
```

I will send a PR.

evilebottnawi:
fixes #256 + tests

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -99,6 +99,8 @@ const Position = require("./ast/position");
  *     - [Block](#block)
  *       - [Program](#program)
  *       - [Namespace](#namespace)
+ *     - [PropertyStatement](#propertystatement)
+ *     - [Property](#property)
  *     - [Declaration](#declaration)
  *       - [Class](#class)
  *       - [Interface](#interface)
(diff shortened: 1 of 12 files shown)


src/ast.js
```diff
@@ -296,8 +297,12 @@ AST.prototype.resolvePrecedence = function(result, parser) {
...
       }
     }
-  } else if (result.kind === "silent" && result.expr.right && !result.expr.parenthesizedExpression) {
-    // overall least precedence 
+  } else if (
+    result.kind === "silent" &&
+    result.expr.right &&
+    !result.expr.parenthesizedExpression
+  ) {
+    // overall least precedence
```
evilebottnawi: Just prettier issue when you run `npm run test`

src/parser/class.js
```diff
@@ -143,20 +145,24 @@ module.exports = {
       function read_variable_declaration() {
         const result = this.node("property");
         this.expect(this.tok.T_VARIABLE);
+        let propName = this.node("identifier");
```
evilebottnawi: use `identifier` for property name

ichiriac: :+1:

PR #267, opened 2019-03-02
Fix regression for empty offsetlookup

czosel:
Hi @ichiriac,

I'm currently updating prettier for PHP to the latest version of the parser, and noticed a regression introduced in https://github.com/glayzzle/php-parser/commit/3f20511d3459a9297c99e75a1c6c86424de87512:

Input: `$var[] = "var"`

Error:
```
TypeError: Cannot read property 'end' of undefined

      at AST.Object.<anonymous>.AST.resolveLocations (node_modules/php-parser/src/ast.js:196:42)
```
This is my attempt to fix it :wink:

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -217,7 +217,7 @@ AST.prototype.resolvePrecedence = function(result, parser) {
   } else if (
     result.kind === "propertylookup" ||
     result.kind === "staticlookup" ||
-    result.kind === "offsetlookup"
+    (result.kind === "offsetlookup" && result.offset)
   ) {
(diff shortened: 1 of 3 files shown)


czosel:
Tests should pass once #268 was merged.

ichiriac:
Hi @czosel, 

Thanks, that's a quick fix but does not really fixes the problem, The script reaches EOF and does not provides the location, it may be related on the way the location is resolved on the last token. I'll merge it and fill an issue to fix the problem at its roots

PR #268, opened 2019-03-02
Update dependencies, fix tests

czosel:


czosel:
@ichiriac It looks like the lockfile hasn't been updated in a very long time. What do you think about adding something like renovate or dependabot to make sure that dependencies always stay up to date?

ichiriac:
yes, great idea, i'll try with dependabot

PR #361, opened 2019-08-27
feat: use `identifier` for goto and label

evilebottnawi:
Parser for php written on PHP do same, also it is allow to use better output comments

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -353,11 +353,15 @@ module.exports = {
       case this.tok.T_STRING: {
         const result = this.node();
         const current = [this.token, this.lexer.getState()];
-        const label = this.text();
+        const labelNameText = this.text();
+        let labelName = this.node("identifier");
         // AST : https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L457
         if (this.next().token === ":") {
+          labelName = labelName(labelNameText);
           this.next();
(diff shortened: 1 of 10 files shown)


test/snapshot/__snapshots__/acid.test.js.snap
```diff
@@ -3118,7 +3118,23 @@ Program {
...
-                "name": "next",
+                "name": Identifier {
+                  "kind": "identifier",
+                  "loc": Location {
+                    "end": Position {
+                      "column": 5,
+                      "line": 83,
+                      "offset": 1665,
+                    },
+                    "source": ":",
```
ichiriac: bad node positions

src/parser/statement.js
```diff
@@ -353,11 +353,13 @@ module.exports = {
       case this.tok.T_STRING: {
         const result = this.node();
         const current = [this.token, this.lexer.getState()];
-        const label = this.text();
+        const labelNameText = this.text();
```
ichiriac: You need to start the node when the text token is not yet finished
```
const labelNameText = this.text();
const labelName = this.node('identifier');
if (this.next().token === ":") {
  labelName = labelName(labelNameText);
  this.next();
  return result("label", labelName);
} else {
  labelName.destroy();
}
```

PR #364, opened 2019-08-27
refactor: exit node

evilebottnawi:
The `status` property is misleading, based on spec (https://github.com/php/php-src/blob/php-7.4.0beta4/Zend/zend_language_parser.y#L1129) it is expression + tests

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -323,18 +323,15 @@ AST.prototype.resolvePrecedence = function(result, parser) {
         result = buffer;
       }
     }
-  } else if (
-    result.kind === "silent" &&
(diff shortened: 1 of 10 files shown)


ichiriac:
Hi @evilebottnawi,

It's not the spec, it's the gammar name, can contain an expression. Here the specification https://www.php.net/manual/en/function.exit.php - the exit argument which is an expression is called the status, in reference to the status code when you exit an program.

I don't understand why it's missleading ?

evilebottnawi:
@ichiriac besed on grammar it is expression, also other parser use `expression` field for this node

PR #365, opened 2019-08-27
test: break and continue

evilebottnawi:
before 5.4.0 it is possible to have `expr` in these nodes (https://github.com/php/php-src/blob/php-7.4.0beta4/Zend/zend_language_parser.y#L437), just test


ichiriac:
thanks

PR #367, opened 2019-08-27
fix: don't add unnecessary `noop`

### LINKED ISSUE #358: still invalid ast for list with multiple `,`
Invalid ast:

Input:
```
list(0 => $var,) = $arr;
```

Should no `noop` element in this case

we need null/`noop` if two or more `,`

evilebottnawi:
fixes #358

No one parsers do it, it is unnecessary node, of somebody can check `trailing comma` exists or not can use `loc` and source code for this

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -323,18 +323,15 @@ AST.prototype.resolvePrecedence = function(result, parser) {
         result = buffer;
       }
     }
-  } else if (
-    result.kind === "silent" &&
-    !result.expr.parenthesizedExpression
-  ) {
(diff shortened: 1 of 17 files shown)


ichiriac:
I would prefer to to exchange ideas before implementing a PR, it may avoid to loose your time.

I'm not rigid about to use or not noop nodes, but I wiil be rigid on the consistency. You can't have this :

```
[$a, $b, ]
// array(var(a), var(b))
[$a, $b, , ]
// array(var(a), var(b), noop)
```

If you use noop, then you use it every time, if you want to ignore it, you don't use it at all.

PR #369, opened 2019-08-27
fix: heredoc and nowdoc

evilebottnawi:
- fix `raw` for `nowdoc` and `heredoc`
- tests


ichiriac:
many thanks @evilebottnawi

PR #376, opened 2019-08-28
test: assign by ref

evilebottnawi:
We should saving test to avoid future regression, it is very bad practice if we don't do it

### DIFF
--- a/test/snapshot/assign.test.js
+++ b/test/snapshot/assign.test.js
@@ -46,4 +46,7 @@ describe('assign', () => {
   it('>>=', () => {
     expect(parser.parseEval("$var >>= $var;")).toMatchSnapshot();
   });
+  it('with ref', () => {
+    expect(parser.parseEval("$bar = &$foo;")).toMatchSnapshot();
+  });
 });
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Please don't merge failed PRs (linting) :disappointed: 

Maintaining the repository is very terrible, let's stick to standard practices, this solved most of the problems

ichiriac:
I think it's simpler to add `--fix` on linter for automatic fixes on precommit hook

test/snapshot/assign.test.js
```diff
@@ -46,4 +46,7 @@ describe('assign', () => {
   it('>>=', () => {
     expect(parser.parseEval("$var >>= $var;")).toMatchSnapshot();
   });
+  it('with ref', () => {
+    expect(parser.parseEval("$bar = &$foo;")).toMatchSnapshot();
```
ichiriac: does not test positions

evilebottnawi: for position we use other test file, i think we should keep position for all tests, but let's do it in other PR

PR #377, opened 2019-08-28
fix: curly offsetlookup

### LINKED ISSUE #374: invalid ast for `{` and `}`
Input:
```php
$var = $foo->bzr_{1};
$var = $foo->bzr_{'string'};
$var = $foo->bzr_{$baz};
$var = $foo->bzr_{$baz->foo};
$var = $foo->bzr_{$var ? 'one' : 'two'};
```

Parsed as encapsed with `offset`, but it is just offsetlookup with curly `brackets`

Should be parse as:
```php
$var = $foo->bzr_[1];
$var = $foo->bzr_['string'];
$var = $foo->bzr_[$baz];
$var = $foo->bzr_[$baz->foo];
$var = $foo->bzr_[$var ? 'one' : 'two'];
```

evilebottnawi:
It is basic tests

I find a lot of check on `[`, but array access can be done with `{`, need investigate

fixes #374

/cc @ichiriac

### DIFF
--- a/src/parser/variable.js
+++ b/src/parser/variable.js
@@ -126,33 +126,6 @@ module.exports = {
         if (is_static_lookup && this.token === this.tok.T_OBJECT_OPERATOR) {
           this.error();
         }
-
-        if (this.token === this.tok.T_VARIABLE) {
-          const inner = this.node("variable");
-          name = this.text().substring(1);
-          this.next();
-          what = this.node("encapsed")(
-            [what, inner(name, false)],
-            null,
-            "offset"
-          );
-          if (what.loc && what.value[0].loc) {
-            what.loc.start = what.value[0].loc.start;
-          }
-        } else if (this.token === "{") {
-          // EncapsedPart
-          const part = this.node("encapsedpart");
-          const expr = this.next().read_expr();
-          this.expect("}") && this.next();
-          what = this.node("encapsed")(
-            [what, part(expr, true)],
-            null,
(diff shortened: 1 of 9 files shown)


src/parser/variable.js
```diff
@@ -209,17 +182,18 @@ module.exports = {
           }
           break;
         case "[":
+        case "{":
           node = this.node("offsetlookup");
           this.next();
           offset = false;
           if (encapsed) {
             offset = this.read_encaps_var_offset();
```
ichiriac: can you check `$obj{$foo]` and `$obj[$bar}` they should fail, but in this implementation they may pass

src/parser/variable.js
```diff
@@ -209,17 +182,18 @@ module.exports = {
           }
           break;
         case "[":
+        case "{":
           node = this.node("offsetlookup");
```
ichiriac: can you add a new `curly` property in order to keep track of offset type of `{` or `[`

evilebottnawi:
:+1: WIP

src/parser/variable.js
evilebottnawi: @ichiriac we have several places with `offsetlookup`, i want to refactor them and test before implement `curly` option (maybe some of them invalid), so let's do it in other PR

ichiriac: ok, lets do it later - it was referenced to #380

PR #379, opened 2019-08-28
fix: string value

### LINKED ISSUE #245: Wrong value when using octal notation, hexadecimal notation and unicode codepoint escaped characters in string
http://php.net/manual/en/language.types.string.php#language.types.string.syntax.double

Test script:

```php
<?php
echo "\n"; // linefeed (LF or 0x0A (10) in ASCII)
echo "\r"; // carriage return (CR or 0x0D (13) in ASCII)
echo "\t"; // horizontal tab (HT or 0x09 (9) in ASCII)
echo "\v"; // vertical tab (VT or 0x0B (11) in ASCII) (since PHP 5.2.5)
echo "\e"; // escape (ESC or 0x1B (27) in ASCII) (since PHP 5.4.4)
echo "\f"; // form feed (FF or 0x0C (12) in ASCII) (since PHP 5.2.5)
echo "\\"; // backslash
echo "\$"; // dollar sign
echo "\""; // double-quote
echo "\141"; // octal notation
echo "\x61"; // hexadecimal notation
echo "\u{0061}"; // Unicode codepoint, which will be output to the string as that codepoint's UTF-8 representation (added in PHP 7.0.0)
```

Normal escape chars `\n\r\t\v\e\f\\\$\"` got expected results.

But octal notation `"\141"`, hexadecimal notation `"\x61"` and unicode codepoint `"\u{0061}"` are not parsed properly. The `raw` is OK, but `value` are not `a`s.

Tested on commit https://github.com/glayzzle/php-parser/blob/64cd69d4cb2bb1bb2348239f2868c95a84a331f8/src/lexer/strings.js#L85-L126

May be related with https://github.com/glayzzle/php-parser/issues/130 https://github.com/glayzzle/php-parser/issues/116

evilebottnawi:
fixes #245

Here interesting place: 
```php
$var = "\u{D83D}\u{DCA9}";
```

Parser on php output: `��` (`strlen` return `6`), because:
> A string is series of characters, where a character is the same as a byte. This means that PHP only supports a 256-character set, and hence does not offer native Unicode support.

But our parser written on `nodejs` and strings will be re-encoded as a UTF-16 string.
So in test we have `value: '💩'`

I think it is expected behavior. In theory we can convert our string to the string is series of characters, but i think it is out of scope parser.

/cc @ichiriac What do you think?

### DIFF
--- a/src/parser/scalar.js
+++ b/src/parser/scalar.js
@@ -6,16 +6,14 @@
 "use strict";
 
 const specialChar = {
-  "\\r": "\r",
-  "\\n": "\n",
-  "\\t": "\t",
-  "\\v": String.fromCharCode(11),
-  "\\e": String.fromCharCode(27),
-  "\\f": String.fromCharCode(12),
-  "\\\\": "\\",
-  "\\$": "$",
-  '\\"': '"',
-  "\\'": "'"
+  "\\": "\\",
+  $: "$",
+  n: "\n",
+  r: "\r",
+  t: "\t",
+  f: String.fromCharCode(12),
+  v: String.fromCharCode(11),
+  e: String.fromCharCode(27)
 };
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


ichiriac:
seems correct to me, just check the changed code, and I'll merge the PR

PR #383, opened 2019-08-30
refactor and test switch statement

evilebottnawi:
- many tests
- refactor
- better error due refactor

### DIFF
--- a/src/parser/switch.js
+++ b/src/parser/switch.js
@@ -42,9 +42,10 @@ module.exports = {
     } else {
       this.expect(["{", ":"]);
     }
+    this.next();
     // OPTIONNAL ';'
     // https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L570
-    if (this.next().token === ";") {
+    if (this.token === ";") {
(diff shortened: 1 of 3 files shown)


src/parser/switch.js
```diff
@@ -34,17 +34,17 @@ module.exports = {
     // DETECT SWITCH MODE
     let expect = null;
     const result = this.node("block");
+    this.expect(["{", ":"]);
```
ichiriac: expect was running before only if token != '{' && != ':' now it runs every time, even on right tokens, no need to call this extra check, make it slower, please can you put as before

src/parser/switch.js
```diff
@@ -65,21 +65,20 @@ module.exports = {
    */
   read_case_list: function(stopToken) {
     const result = this.node("case");
+    this.expect([this.tok.T_CASE, this.tok.T_DEFAULT]);
```
ichiriac: same, less effective

evilebottnawi:
@ichiriac fixed :+1:

PR #388, opened 2019-08-31
refactor: byRef

evilebottnawi:
/cc @ichiriac here just fix

in the next PR i want to remove `byRef` node and revert some changes from https://github.com/glayzzle/php-parser/commit/fbfd46f7401c468ef38eac9e50b5ac8ed35cde9c

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -46,6 +46,7 @@ const Position = require("./ast/position");
  *     - [Exit](#exit)
  *     - [Clone](#clone)
  *     - [Assign](#assign)
+ *     - [AssignRef](#assignref)
  *     - [Array](#array)
  *     - [List](#list)
  *     - [Variable](#variable)
@@ -458,6 +459,7 @@ AST.prototype.checkNodes = function() {
 [
   require("./ast/array"),
   require("./ast/assign"),
+  require("./ast/assignref"),
   require("./ast/bin"),
   require("./ast/block"),
   require("./ast/boolean"),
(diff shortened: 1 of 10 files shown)


evilebottnawi:
@ichiriac hm, why? `AssignRef` === `Assign` with `ref`, we can reduce unnecessary nodes, php parser has many unnecessary nodes

ichiriac:
`AssignRef` !== `Assign` - because `AssignRef` works only on `=` operators, so it's a subset of assign (without the need of operator property), also `AssignRef` does not have the same precedence as `Assign` with operator `=`

evilebottnawi:
@ichiriac after `byRef` node, a lot of code very dirty, it is hard to do in one PR, maybe we can merge this as is and i will do this in other PR, also i can't find information about `precedence` https://www.php.net/manual/en/language.operators.precedence.php

evilebottnawi:
ok, WIP

evilebottnawi:
@ichiriac done :+1:

ichiriac:
thanks @evilebottnawi - it's great

evilebottnawi:
@ichiriac WIP on remove unnecessary stuff for `byRef` node

ichiriac:
in principle I don't use byRef node anymore, see https://github.com/glayzzle/php-parser/commit/72cde999df66069aee8ca3f58b4147f7df4090cc - but still it may be cases where the byref syntax is too permissive

evilebottnawi:
@ichiriac i will check all :+1:

PR #392, opened 2019-08-31
refactor: read_internal_functions_in_yacc

evilebottnawi:
Just refactor for better reading and simplify for better maintenance 

In some places we can see unnecessary (extra) checks, but it seems to me that support is more important than speed at the moment

### DIFF
--- a/src/parser/expr.js
+++ b/src/parser/expr.js
@@ -126,8 +126,70 @@ module.exports = {
   /**
    * Reads isset variables
    */
-  read_isset_variables: function () {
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@ichiriac in `acorn` :smile: , when you use over `+`/`-`/etc 9k you got error, they fixed this, but i think it is not high priority right now, I would like to stabilize the parser at the beginning and then correct such errors, because they are extremely rare in current code, maybe we can merge this as is and i will investigate this in future? We can open issue about this

PR #402, opened 2019-09-11
fix: invalid parsing of traitalias

evilebottnawi:
- Refactor code using `read_*` for better maintenance
- Fix parsing:
```php
class MyClass1 {
    use HelloWorld { sayHello as abstract private test; }
}
```

It is invalid and should not be parsable.

- Tests

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -203,6 +203,40 @@ module.exports = {
 
     return result(null, items, flags);
   },
+
+  read_member_modifier: function() {
+    let modifier;
+
+    switch (this.token) {
+      case this.tok.T_PUBLIC:
+        modifier = 0;
+        break;
+      case this.tok.T_PROTECTED:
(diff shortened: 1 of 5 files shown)


src/parser/class.js
```diff
@@ -213,31 +247,34 @@ module.exports = {
...
-        val = 0;
       do {
-        switch (this.token) {
-          case this.tok.T_PUBLIC:
+        let idx = 0;
+        let val = 0;
+
+        const visibility = this.read_member_modifier();
+
+        switch (visibility) {
```
evilebottnawi: Need refactor too, let's do it in next PRs

ichiriac:
Hi @evilebottnawi, not sure what to do with this PR, can we review it on 3.1 or does it introduce breaking changes ?

evilebottnawi:
Yes, let'd o it in `3.1`

PR #407, opened 2019-09-11
Feat spread operator for array

evilebottnawi:
- support spread operator (`...`) for array
- always use `entry` for any item in array (PHP parser do same)
- add `byRef` to `entry` node if we use `&` with `variable` to avoid invalid grammatical
- add `unpack` when `...` uses
- tests
- remove unnecessary `read_byref` due refactor

Entry when `key` is `null` means value item (`[$var, $var2, $var3]`)
Entry when `key` is not `null` means key-value item (`['one' => $var, 'two' => $var, 'three' => $var]`)

Interesting note - grammar support `...` in `list` and in `foreach`, but php throw error in runtime, we need do same in future, now it is impossible due invalid grammar in `read_expr_item`, we need refactor and them do it, it is not high priority because nobody writes this code


ichiriac:
@evilebottnawi does this branch should be merges ?

evilebottnawi:
Need review and we can merge this :smile:

PR #444, opened 2019-12-06
Add support for PHP RFC: Flexible Heredoc and Nowdoc Syntaxes

rainx:
This update one jest snapshot file, please check that. let me know if it needs some changes.

### DIFF
--- a/src/lexer/strings.js
+++ b/src/lexer/strings.js
@@ -65,7 +65,8 @@ module.exports = {
             this._input[this.offset - 1] === "\n"
           ) {
             // go go go
-            this.heredoc_label = yylabel;
+            this.heredoc_label.label = yylabel;
+            this.heredoc_label.length = yylabel.length;
             yyoffset = this.offset - revert;
             this.offset = revert;
             this.consume(yyoffset);
@@ -74,6 +75,8 @@ module.exports = {
             } else {
               this.begin("ST_HEREDOC");
             }
+            // prematch to get the indentation information from end of doc
+            this.prematch_ENDOFDOC();
             return this.tok.T_START_HEREDOC;
           }
         }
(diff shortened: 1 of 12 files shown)


rainx:
> Please add tests

Added

src/lexer/strings.js
```diff
@@ -126,25 +129,100 @@ module.exports = {
...
+    while (/[\t ]/.test(leading_ch)) {
+      if (leading_ch === " ") {
+        indentation_uses_spaces = true;
+      } else if (leading_ch === "\t") {
+        indentation_uses_tabs = true;
+      }
+
+      leading_ch = this._input[offset + indentation];
+      indentation++;
+    }
```
evilebottnawi: I think we should throw error on invalid indentation like:
```
// Parse error: Invalid body indentation level (expecting an indentation at least 5) in %s on line %d
```

How in rfc

rainx: Get it. I will added it later

rainx: > I think we should throw error on invalid indentation like:
> 
> ```
> // Parse error: Invalid body indentation level (expecting an indentation at least 5) in %s on line %d
> ```
> 
> How in rfc

Just added code to detect this kind of error and raise the error via `this.raiseError` , I am not sure if I handle it in the right way. please help check it.

evilebottnawi:
/cc @ichiriac need review too, thanks

for me looks good

src/lexer/strings.js
```diff
@@ -126,25 +129,100 @@ module.exports = {
...
+
+    // skip leading spaces or tabs
+    let indentation_uses_spaces = false;
+    let indentation_uses_tabs = false;
+    // reset heredoc_label structure
+    let indentation = 0;
+    let leading_ch = this._input.substring(offset - 1, offset);
+
+    while (/[\t ]/.test(leading_ch)) {
```
ichiriac: please, no regex as they perform about 10 times slower than simple char checks `while ( leading_ch == '\t' || leading_ch == ' ') { `

ichiriac:
@evilebottnawi I will finish the review tomorrow

src/lexer/strings.js
rainx: Ok.. just removed this regex in loop

rainx:
> @evilebottnawi I will finish the review tomorrow

Any other updates?

src/parser/scalar.js
```diff
@@ -79,6 +86,65 @@ module.exports = {
...
+      } else {
+        inCoutingState = false;
+      }
+
+      if (inCheckState && leadingWhitespaceCharCount < indentation) {
+        this.raiseError(
+          `Invalid body indentation level (expecting an indentation at least ${indentation})`
+        );
+      } else {
+        inCheckState = false;
```
ichiriac: Be aware that `raiseError` will not throw an error in `suppressErrors` mode. You shoud avoid else as if the process continues it should fallback in a normal state

src/lexer/strings.js
ichiriac: It's ok, but you should also test the code with `suppressErrors` mode in order to check it's fallback mode

ichiriac:
It's checked and it's all good, thanks for the hard work and sorry for the late reply. I'll make some changes and then release it under a stable version.

Many thanks again for your contribution !

PR #455, opened 2020-01-01
fix: silent precendence

### LINKED ISSUE #195: [feature] Convert silent mode to a property
Same idea as `parenthesizedExpression`, use a new property that replace the `silent` node with a `silent` property (true/undefined) in order to reprensent the `@` symbol.

Related issues : #177 and #170

### LINKED ISSUE #355: invalid ast for silent
Input:
```php
<?php

echo '<select name="foo"><option value="bar"' . (@$_GET['foo'] === 'bar' ? ' 
 selected="selected"' : '') . '>bar</option></select>';
```

Expected: the `silent` node should be as `left` property of `bin`, now `silent` under `retif`

php parser output (https://github.com/nikic/PHP-Parser):
```
array(
    0: Stmt_Echo(
        exprs: array(
            0: Expr_BinaryOp_Concat(
                left: Expr_BinaryOp_Concat(
                    left: Scalar_String(
                        value: <select name="foo"><option value="bar"
                    )
                    right: Expr_Ternary(
                        cond: Expr_BinaryOp_Identical(
                            left: Expr_ErrorSuppress(
                                expr: Expr_ArrayDimFetch(
                                    var: Expr_Variable(
                                        name: _GET
                                    )
                                    dim: Scalar_String(
                                        value: foo
                                    )
                                )
                            )
                            right: Scalar_String(
                                value: bar
                            )
                        )
                        if: Scalar_String(
                            value:  
                         selected="selected"
                        )
                        else: Scalar_String(
                            value: 
                        )
                    )
                )
                right: Scalar_String(
                    value: >bar</option></select>
                )
            )
        )
    )
)
```

`Expr_ErrorSuppress` === `silent`

czosel:
Change `silent` to be handled exactly the same way as `cast` is handled. Also refactor `shouldBeSame` for better error reporting.

Fixes #195
Fixes #355

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -253,26 +253,22 @@ AST.prototype.resolvePrecedence = function(result, parser) {
       }
     }
   } else if (
-    (result.kind === "cast" &&
-      result.what &&
-      !result.what.parenthesizedExpression) ||
-    (result.kind === "silent" &&
-      result.expr &&
-      !result.expr.parenthesizedExpression)
+    ["silent", "cast"].includes(result.kind) &&
+    result.expr &&
+    !result.expr.parenthesizedExpression
   ) {
     // https://github.com/glayzzle/php-parser/issues/172
-    const subject = result.kind === "cast" ? "what" : "expr";
-    if (result[subject].kind === "bin") {
-      buffer = result[subject];
-      result[subject] = result[subject].left;
-      this.swapLocations(result, result, result[subject], parser);
+    if (result.expr.kind === "bin") {
+      buffer = result.expr;
+      result.expr = result.expr.left;
+      this.swapLocations(result, result, result.expr, parser);
       buffer.left = this.resolvePrecedence(result, parser);
       this.swapLocations(buffer, buffer.left, buffer.right, parser);
       result = buffer;
-    } else if (result[subject].kind === "retif") {
-      buffer = result[subject];
-      result[subject] = result[subject].test;
-      this.swapLocations(result, result, result[subject], parser);
+    } else if (result.expr.kind === "retif") {
+      buffer = result.expr;
+      result.expr = result.expr.test;
+      this.swapLocations(result, result, result.expr, parser);
       buffer.test = this.resolvePrecedence(result, parser);
       this.swapLocations(buffer, buffer.test, buffer.falseExpr, parser);
       result = buffer;
--- a/src/ast.js
+++ b/src/ast.js
@@ -163,8 +163,7 @@ AST.precedence = {};
   ["*", "/", "%"],
   ["!"],
   ["instanceof"],
-  ["cast"]
-  // TODO: typecasts
+  ["cast", "silent"]
   // TODO: [ (array)
   // TODO: clone, new
 ].forEach(function(list, index) {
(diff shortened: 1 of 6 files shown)


src/ast.js
```diff
@@ -254,22 +253,26 @@ AST.prototype.resolvePrecedence = function(result, parser) {
...
+    (result.kind === "silent" &&
+      result.expr &&
+      !result.expr.parenthesizedExpression)
   ) {
     // https://github.com/glayzzle/php-parser/issues/172
-    if (result.what.kind === "bin") {
-      buffer = result.what;
-      result.what = result.what.left;
-      this.swapLocations(result, result, result.what, parser);
+    const subject = result.kind === "cast" ? "what" : "expr";
```
ichiriac: could be great to rename `what` property to `expr` on `cast` nodes, that way would be more consistent grammatically speaking - the next release will be stable so it's the last chance to introduce breaking changes.

src/ast.js
```diff
@@ -254,22 +253,26 @@ AST.prototype.resolvePrecedence = function(result, parser) {
...
+      result[subject] = result[subject].left;
+      this.swapLocations(result, result, result[subject], parser);
       buffer.left = this.resolvePrecedence(result, parser);
       this.swapLocations(buffer, buffer.left, buffer.right, parser);
       result = buffer;
-    } else if (result.what.kind === "retif") {
-      buffer = result.what;
-      result.what = result.what.test;
-      this.swapLocations(result, result, result.what, parser);
+    } else if (result[subject].kind === "retif") {
```
ichiriac: the last `...if (result.what.kind === "retif"` becomes dead code and should be removed : 
https://coveralls.io/builds/27873326/source?filename=src/ast.js#L291

src/ast.js
czosel: I was hoping you'd suggest that :wink: done!

src/ast.js
czosel: I removed it but I must admit that I don't really see how my change caused this. Can you explain it?

src/ast.js
```diff
@@ -253,26 +253,22 @@ AST.prototype.resolvePrecedence = function(result, parser) {
       }
     }
   } else if (
-    (result.kind === "cast" &&
-      result.what &&
-      !result.what.parenthesizedExpression) ||
-    (result.kind === "silent" &&
-      result.expr &&
-      !result.expr.parenthesizedExpression)
+    ["silent", "cast"].includes(result.kind) &&
```
ichiriac: sorry, I'm a bit maniac, could you rewrite the condition to be more optimized as this function is runned over all nodes it may impact a bit :

`(result.kind == 'silent' || result.kind == 'cast') && ...`

NB : The property check costs less than a function call and a lookup because you have only 2 entries

src/ast.js
ichiriac: before there was duplicated code between `what` and `expr` properties, as your code is more generic here : https://coveralls.io/builds/27873326/source?filename=src/ast.js#L272 the last elseif check does never trigger

src/ast.js
czosel: Fixed! Might be a good idea to add performance testing in the future :wink:

src/ast.js
czosel: Alright! :+1:

src/ast.js
ichiriac: it has something similar :

```
 PASS  test/perf.test.js
  ● Console

    console.log test/perf.test.js:42
          + Tokens speed  => 205.15K/sec
    console.log test/perf.test.js:56
          + Reading speed => 437.1K/sec
    console.log test/perf.test.js:101
          + Nodes speed  => 79.69K/sec
    console.log test/perf.test.js:115
          + Overall speed => 211.43K/sec
```
It's not kilobytes but thousands of tokens or nodes dependending on what is tested.

The problem is that depends on the CPU & RAM frequency (disk does not enter into account) so you can't really know automatically it you break/degrade something - any idea is welcome

czosel: Nice! I have no experience with performance testing myself, but I saw that `prettier/prettier` is doing [something similar](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md#performance) based on [Benchmark.js](https://www.npmjs.com/package/benchmark).

czosel:
@ichiriac What do you think, is this ready to be merged? :rocket:

Issue #57, opened 2017-01-16
improve parsing speed


ichiriac:
The lexer speed (reading a string from memory) is about 2,5 millions tokens / sec (depends on CPU).

It's about 3 times more slower than original PHP algorithm, but about 16 times more rapid than a jison based version (which is also more rapid than a regex version ...)

The lexer algorithm could be improved a bit, but there will not be much gain here.

The parser sometimes is 10x slower than the lexer, there is a lot of room to improve the speed, the most cost effective part is recursive parsing on structures like arrays, property chains or string concatenations.

ichiriac:
List of lexer optimizations :

- [ ] Avoid costs with unput calls, instead use of appendToken in order to improve speeds
- [ ] Check if need recusivity/stacks on lexing states and avoid dynamic calls if possible (boolean states)
- [ ] Try to improve the unput function (columns problem)

List of parseroptimizations :

- [ ] Convert expression recursivity into a loop (avoid huge stacks)
- [ ] Add benchmarks & profiling

ichiriac:
- [ ] Replace array lookups with maps : https://jsperf.com/object-hasownproperty-vs-array-indexof/39
- [ ] Add output on tests suite in order to keep track of lexing and parsing speed

PR #458, opened 2020-01-02
start to track performance evolution

ichiriac:
related to a discussion on #455 - @czosel you are welcome to contribute on this :

- [ ] We could use benchmark.js framework as the results may be more accurate
- [ ] Add a traking JSON `test/perf/statistics.json`
- [ ] At each run lookup if there is a result entry
- [ ] Scan the `test/perf` folder for php files, and on each run tests on tokens & nodes (and check the threshold)
- [ ] If entry found, check with the threshold if the performance is ok
- [ ] If no entry, at the end of tests, create one and update the JSON file
- [ ] Add more tests to cover various use cases (but keep them generic enough to be run on php7.0)
- [ ] Checkout previous release tags and try to get performance statistics on each in order to see if performance has been degraded

### DIFF
--- a/test/perf.test.js
+++ b/test/perf.test.js
@@ -1,5 +1,7 @@
-var parser = require("./main");
-
+const parser = require("./main");
+const os = require("os");
+const cpus = os.cpus();
+const cpuId = cpus.length + "@" + cpus[0].speed + 'Mhz';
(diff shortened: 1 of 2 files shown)


ichiriac:
related to #57

czosel:
Landed in #885

PR #459, opened 2020-01-02
feat: lint tests as well

czosel:


czosel:
We don't have to merge this yet if it's causing too many conflicts with existing PRs. But I think it's a good idea to enable linting for tests as well, because it helps improve their quality and enables auto-formatting.

PR #460, opened 2020-01-02
fix: regression with unary / retif precedence

czosel:
I noticed during the update of prettier/plugin-php that we shouldn't have removed the code block handling unary / retif in #455 :wink: 
This reverts the code block and adds a test.


test/precedence.test.js
```diff
@@ -158,4 +158,8 @@ describe("Test precedence", function() {
     shouldBeSame("$a = (string)$b->foo . $c", "$a = ((string)$b->foo) . $c");
     shouldBeSame("(bool) $var ? 1 : 2;", "((bool)$var) ? 1 : 2;");
   });
+
+  it("test unary / retif", function() {
+    shouldBeSame("$a = +(+$var ? 1 : 2)", "$a = +((+$var) ? 1 : 2)")
```
ichiriac: :+1:

PR #463, opened 2020-01-03
chore: refactor testing infrastructure

czosel:
* extract jest config into jest.config.js
* run eslint using jest-runner-eslint
* rename "lint" script to "fix" because it includes --fix
* remove "cover" script, coverage can still be tested locally with
`COVERAGE=1 npm test`
* add eslint-plugin-jest
* remove some redundant tests
* refactor some tests using it.each

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ notifications:
     on_success: change
     on_failure: always
     on_start: never
-script: npm run lint && npm run cover
+script: npm test
 after_success: cat /home/travis/build/glayzzle/php-parser/coverage/lcov.info | /home/travis/build/glayzzle/php-parser/node_modules/coveralls/bin/coveralls.js
 after_script:
   - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,20 @@
+"use strict";
+
+const ENABLE_COVERAGE = !!process.env.CI || !!process.env.COVERAGE;
+
+module.exports = {
+  collectCoverage: ENABLE_COVERAGE,
+  coverageDirectory: "coverage/",
+  projects: [
+    {
+      displayName: "test",
+      testEnvironment: "node"
+    },
+    {
+      runner: "jest-runner-eslint",
+      displayName: "lint",
+      testMatch: ["<rootDir>/**/*.js"],
+      testPathIgnorePatterns: ["<rootDir>/node_modules/", "<rootDir>/coverage/"]
+    }
+  ]
+};
--- a/package.json
+++ b/package.json
@@ -11,21 +11,18 @@
     "LICENSE"
   ],
   "scripts": {
-    "lint": "eslint . --fix",
+    "fix": "eslint . --fix",
     "test": "jest",
-    "prehusky": "npm run lint",
     "husky": "npm run test",
     "prebuild": "npm run test",
     "build": "webpack --config webpack.config.js",
     "postbuild": "npm run build-docs",
     "build-docs": "jsdoc -c .jsdoc.json",
-    "publish-docs": "git subtree push --prefix docs origin gh-pages",
-    "cover": "jest --coverage --coverageDirectory=coverage/"
+    "publish-docs": "git subtree push --prefix docs origin gh-pages"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run husky",
-      "pre-push": "npm run husky"
+      "pre-commit": "npm run husky"
     }
   },
   "repository": {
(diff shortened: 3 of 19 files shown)


czosel:
I was glad I could use a vim macro to do this :sweat_smile:

czosel:
Up for discussion: running the tests pre-commit _and_ pre-push seems a little bit too much. Shall wen remove pre-push?

ichiriac:
@czosel agree with you, we can remove pre-push

.travis.yml
```diff
@@ -17,7 +17,7 @@ notifications:
     on_success: change
     on_failure: always
     on_start: never
-script: npm run lint && npm run cover
+script: npm test
```
ichiriac: the coverage is automaticlaly pushed by travis to coveralls, should be `npm run cover` instead to continue updating coverage stats

package.json
```diff
@@ -11,16 +11,14 @@
     "LICENSE"
   ],
   "scripts": {
-    "lint": "eslint . --fix",
+    "fix": "eslint . --fix",
     "test": "jest",
-    "prehusky": "npm run lint",
```
ichiriac: actually linting is automatically done before each commit with this command

package.json
```diff
@@ -11,16 +11,14 @@
...
     "husky": "npm run test",
     "prebuild": "npm run test",
     "build": "webpack --config webpack.config.js",
     "postbuild": "npm run build-docs",
     "build-docs": "jsdoc -c .jsdoc.json",
-    "publish-docs": "git subtree push --prefix docs origin gh-pages",
-    "cover": "jest --coverage --coverageDirectory=coverage/"
```
ichiriac: need to keep it because it's used on coveralls

test/snapshot/arrowfunc.test.js
```diff
@@ -1,32 +1,17 @@
...
-    ).toMatchSnapshot();
-  });
-  it("return type", () => {
-    expect(
-      parser.parseEval('$var = fn(): ?string => "something";')
-    ).toMatchSnapshot();
-  });
-  it("inside call", () => {
-    expect(parser.parseEval(`call(fn($arg) => $arg);`)).toMatchSnapshot();
+  it.each([
```
ichiriac: :+1:

test/snapshot/encapsed.test.js
```diff
@@ -1,198 +1,103 @@
...
-  });
-  it("curly #3", () => {
-    expect(parser.parseEval('"string {$call()} string";')).toMatchSnapshot();
-  });
-  it("no curly", () => {
-    expect(parser.parseEval('"string $$juice string";')).toMatchSnapshot();
-  });
-  it("propertylookup", () => {
-    expect(parser.parseEval('$this->{"set{$type}"};')).toMatchSnapshot();
+  it.each([
```
ichiriac: :+1:

test/snapshot/byref.test.js
```diff
@@ -33,9 +33,6 @@ describe("byref", () => {
   it("function definition", () => {
     expect(parser.parseEval("function &foo( &$bar ) { }")).toMatchSnapshot();
   });
-  it("variable", () => {
```
ichiriac: why removing this test ?

test/snapshot/byref.test.js
```diff
@@ -113,11 +110,6 @@ describe("byref", () => {
   it.skip("propertylookup #2", () => {
     expect(parser.parseEval("$var = &($var)->test;")).toMatchSnapshot();
   });
-  it("closure", () => {
```
ichiriac: why removed ?

package.json
czosel: "lint" is now part of "test" which still runs on each commit ("husky")

.travis.yml
czosel: This should still work, it's part of `jest.config.js`

test/snapshot/byref.test.js
czosel: the jest-eslint-plugin revealed that it's a duplicate.

test/snapshot/byref.test.js
czosel: same here

jest.config.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+const ENABLE_COVERAGE = !!process.env.CI || !!process.env.COVERAGE;
+
+module.exports = {
+  collectCoverage: ENABLE_COVERAGE,
+  coverageDirectory: "coverage/",
```
czosel: @ichiriac this is how coverage still works even without `npm run cover`. Try running `COVERAGE=1 npm test`. In CI it works because the `CI` env variable is set.

.travis.yml
czosel: You can also see the coverage still being reported here: https://travis-ci.org/glayzzle/php-parser/builds/632443880?utm_source=github_status&utm_medium=notification

czosel:
@ichiriac The pre-push script is now removed :+1:

czosel:
@ichiriac let me know if there are any more open questions - would be nice if we can get this merged before we get merge conflicts.

jest.config.js
ichiriac: great, didn't knew about that 👍

.travis.yml
ichiriac: great 👍

package.json
ichiriac: 👍

ichiriac:
@czosel you're right, it's ok for me

PR #464, opened 2020-01-03
WIP: Fix function called "fn"

czosel:
Seems like since arrow function support was merged, functions called "fn" can't be parsed anymore.

Based on #463


czosel:
:see_no_evil: Guess I should have read about how arrow functions work in 7.4 first

PR #469, opened 2020-01-08
fix: remove trailing newline in flexible heredoc/nowdoc value

### LINKED ISSUE #468: Flexible nowdoc contains extra linebreak
"Classic" nowdoc nodes contain no trailing linebreak in `value`, while flexible ones do.

```php
echo <<<'EOD'
Example of string
spanning multiple lines
using heredoc syntax.
EOD;
```

```
        Nowdoc {
          "kind": "nowdoc",
          "label": "EOD",
          "raw": "<<<'EOD'
Example of string
spanning multiple lines
using heredoc syntax.
EOD",
          "value": "Example of string
spanning multiple lines
using heredoc syntax.",
        },
```

Input:
```php
echo <<<'EOD'
  Example of string
  spanning multiple lines
  using heredoc syntax.
  EOD;

```

```
        Nowdoc {
          "kind": "nowdoc",
          "label": "EOD",
          "raw": "<<<'EOD'
  Example of string
  spanning multiple lines
  using heredoc syntax.
  EOD",
          "value": "Example of string
spanning multiple lines
using heredoc syntax.
",
        },
```

czosel:
Fixes #468

### DIFF
--- a/src/parser/scalar.js
+++ b/src/parser/scalar.js
@@ -223,6 +223,12 @@ module.exports = {
             node = this.node("nowdoc");
             value = this.next().text();
             // strip the last line return char
+            if (this.lexer.heredoc_label.indentation > 0) {
+              value = value.substring(
+                0,
+                value.length - this.lexer.heredoc_label.indentation
+              );
+            }
             const lastCh = value[value.length - 1];
             if (lastCh === "\n") {
               if (value[value.length - 2] === "\r") {
@@ -431,6 +437,26 @@ module.exports = {
     // reading encapsed parts
     while (this.token !== expect && this.token !== this.EOF) {
       value.push(this.read_encapsed_string_item(true));
+    }
+    if (
+      value.length > 0 &&
+      value[value.length - 1].kind === "encapsedpart" &&
+      value[value.length - 1].expression.kind === "string"
+    ) {
+      const node = value[value.length - 1].expression;
+      const lastCh = node.value[node.value.length - 1];
+      if (lastCh === "\n") {
+        if (node.value[node.value.length - 2] === "\r") {
+          // windows style
+          node.value = node.value.substring(0, node.value.length - 2);
+        } else {
+          // linux style
+          node.value = node.value.substring(0, node.value.length - 1);
+        }
+      } else if (lastCh === "\r") {
+        // mac style
+        node.value = node.value.substring(0, node.value.length - 1);
+      }
     }
     this.expect(expect) && this.next();
     const raw = this.lexer._input.substring(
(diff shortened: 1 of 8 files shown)


ichiriac:
Hi @czosel, problem comes from here : https://github.com/glayzzle/php-parser/blob/master/src/parser/scalar.js#L227

In this function I remove the last line break, but here the last char is the indentation, so you should before checking add an extra check on indentation, remove the trailing indentation, and then keep the strip of last line break.

czosel:
Thanks for the pointer @ichiriac! I added a fix for nowdoc, but heredoc now I still have to figure out heredoc. Do you have another pointer? :wink:

src/parser/scalar.js
```diff
@@ -223,19 +223,7 @@ module.exports = {
...
-                value = value.substring(0, value.length - 2);
-              } else {
-                // linux style
-                value = value.substring(0, value.length - 1);
-              }
-            } else if (lastCh === "\r") {
-              // mac style
-              value = value.substring(0, value.length - 1);
-            }
+            value = value.replace(/(\r\n?|\n) *$/, "");
```
ichiriac: not 100% sure but I believe that's not only spaces, could be tabs also. In order to avoid extra regex call could you do something like that :

```
if (this.lexer.heredoc_label.indentation > 0) {
  value = value.substring(0, value.length - this.lexer.heredoc_label.indentation;
}
// and keep the rest the same 
const lastCh = value[value.length - 1];
```

ichiriac:
Not sure, but from reading the code it may be handled here : https://github.com/glayzzle/php-parser/blob/master/src/parser/scalar.js#L329

```
From case this.tok.T_START_HEREDOC: ->
if (this.lexer.curCondition === "ST_NOWDOC") {
} else {
--> return this.read_encapsed_string(this.tok.T_END_HEREDOC);
enter into a loop :
while (this.token !== expect && this.token !== this.EOF) {
      value.push(this.read_encapsed_string_item(true));
    }
and then using if (this.token === this.tok.T_ENCAPSED_AND_WHITESPACE) 
```

The problem here is that you may also have tokens T_ENCAPSED_AND_WHITESPACE at the start;, middle, and at the end ... so you should include the fix here : https://github.com/glayzzle/php-parser/blob/master/src/parser/scalar.js#L435

By testing if the last node kind is a `string` and something like this :

 ```
if (value.length > 0 && value[value.length -1].kind === 'string') {
  const node = value[value.length -1];
  if ( this.lexer.heredoc_label.indentation > 0) {
      node.value = node.value.substring(0, node.value.length - this.lexer.heredoc_label.indentation;
  }
  // because it was never handled before
  const lastCh = ... 
}
```

czosel:
Thanks again for the pointers @ichiriac - I made the changes. I didn't need to remove the indentation for heredoc, because that's already done in `remeve_heredoc_leading_whitespace_chars`.

PR #473, opened 2020-01-12
null coalesce operator should be right-associative

czosel:
This is actually just a bug report with an attached test: The null coalescing operator is currently evaluated in a left-associative style, while it should be right-associative.

### DIFF
--- a/test/snapshot/__snapshots__/bin.test.js.snap
+++ b/test/snapshot/__snapshots__/bin.test.js.snap
@@ -561,6 +561,41 @@ Program {
 }
 `;
 
+exports[`bin ?? right-associative 1`] = `
+Program {
+  "children": Array [
+    ExpressionStatement {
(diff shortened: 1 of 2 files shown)


czosel:
@ichiriac Shall I add an issue for this instead? The test by itself just shows that this is not correct right now.

ichiriac:
oups, sorry thought you just added a new test, can you make a new PR on this ?

czosel:
@ichiriac I can give it a shot - do you have a pointer for me on how associativity of operators is handled?

ichiriac:
it's on ast.js in resolvePrecedence function, you need to identify when the node needs to change associativity, and then change properties, and use swapLocations to update their position.

It's not a prority for 3.0 stable release, can be released as a minor / bugfix

czosel:
Thanks for the pointer! See #485

Issue #362, opened 2019-08-27
rename `classreference` to `name` node


evilebottnawi:
Why? `classreference` is very misleading, in `php` we have `name` (https://www.php.net/manual/en/language.namespaces.rules.php), also php parser on php and go use `name` node for this (they have `Name`, `Name_FullyQualified` and `Name_Relative` nodes, but it is unnecessary ahve a lot of nodes when we have `resolution` property).

Also php use `name` in source `https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L307`

I can send a PR

Also i find some nodes use `identifier` instead `classreference`(`name`), we should fix it, i will send PRs

ichiriac:
Hi @evilebottnawi, your PR is welcome for renaming classreference to name if you feel it more intuitive.

For identifier, it's not so simple. That's maybe why I find classreference less confusing than name

Some examples :

```php
class foo extends bar {
}
$a = new foo(); 
```

The class name is an identifier, like function names, constants names. An identifier is just a normalized identifier name. The `bar` on extends is a class reference. It could be a name, but it has extra information about namespace, identifiers can't.

The two are not the same. On JS for example they use identifiers, and don't have a name equivalent. I want to keep the difference, and as usual PHP make bad choices on mix concepts (https://eev.ee/blog/2012/04/09/php-a-fractal-of-bad-design/)

Classreference is not the good name anyway, as it can be n interface, constant, function name, so we need a better term - I'm open on any suggestions

evilebottnawi:
@ichiriac hm, maybe we should look at the grammar and use and do it based on the grammar, maybe we need more nodes, i will write about new nodes/rename nodes and change some nodes in nodes here in near future

ichiriac:
Ok, then, identifier still remains the same :
https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L288

And here an usage example : 
https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1121

For  `name`, here the definition :
https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L307

And here an usage :
https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L1053

The name tag is related exclusively namespace resolution and usage by various nodes like functions, classes, constants, etc ... so classreference is a bad choice - I'm ok to rename it name as it's generic.

It would be nice if you can do a PR on this change

evilebottnawi:
[...]

PR #478, opened 2020-01-13
Classreference to name

ichiriac:
impl of #362 - @czosel / @evilebottnawi - this change may also introduce changes on prettier, could you check if it's ok before merging ?


czosel:
The rename is no problem for prettier. I can't really say if "name" is a better node name than "classreference" though.

Issue #511, opened 2020-03-21
Invalid start location for class constants


czosel:
Input:
```php
<?php
class A
{
    const test = 1;
}
```
AST (excerpt):
```
"kind": "classconstant",
          "loc": {
            "source": "test = 1",
            "start": {
              "line": 4,
              "column": 10,
              "offset": 26
            },
            "end": {
              "line": 4,
              "column": 18,
              "offset": 34
            }
          },
```
I think the start column of the class constant is off, it should be around 4 instead of 10 in the example.

czosel:
The same thing happens with class properties:
```php
<?php
class A
{
    public $test = 1;
}
```
AST:
```
"kind": "propertystatement",
          "loc": {
            "source": "$test = 1",
            "start": {
              "line": 4,
              "column": 11,
              "offset": 27
            },
            "end": {
              "line": 4,
              "column": 20,
              "offset": 36
            }
          },
```

evilebottnawi:
Yes, it is old issues :disappointed:

czosel:
Maybe related to https://github.com/glayzzle/php-parser/issues/208 ?

evilebottnawi:
Yes, i try to keep all in the one PR, but it's just my preference

czosel:
Doesn't matter as long as we get it fixed :wink:

ichiriac:
for const the fix is easy, but for properties / functions it's more complicated as the final / abstract / private / protected / public tokens may be eaten before creating the function or property node - one solution would be to use swapLocations on the generated and store on flags the parser position before  eating modifiers tokens

ichiriac:
Hi @evilebottnawi, it would be great if you have some time that you keep working on branch #fix-issue-511 as the fix is not fully implemented. 

Remains to check interfaces, traits, use alias on traits, but the principle remains the same as in the last commit

ctf0:
for the class props, i used this to fix the range size, there are some cases this will fail, but for the correctly formated it will work as expected

```ts
const _set = require('lodash.set');

_classAST?.body
  .filter((item: any) => item.kind == 'propertystatement')
  .map((item: any) => {
      const start = item.loc.start
      let extraLength = start.column - (item.visibility.length + 1) // +1 > include the space after 'visibility'

      if (item.isStatic) {
          extraLength -= 'static '.length
      }

      _set(item, 'loc.start.column', extraLength)
[...]

PR #549, opened 2020-05-10
fix(precedence): right-associative operators

### LINKED ISSUE #548: Precedence of exponentiation operator
Hi @ichiriac 

In the PHP plugin for prettier an issue with the precedence of the exponentiation operator (`**`) was reported (https://github.com/prettier/plugin-php/issues/1421):

Input:
```php
1 + 2 ** 3;
1 ** 2 + 3;
```

The first expression is parsed incorrectly as `(1+2)**3`, the second correctly as `(1**2)+3` ([see playground](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDCICM7azDBBCbwQAB4AfmMwAB1MbYBGAAIAan3s-YAqM-2AZgBuPfPL0+Pb7ZB2oA))

Screenshot from 2020-05-10 13-07-02

czosel:
This fixes a regression introduced in #485 

Fixes #548

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -239,7 +239,8 @@ AST.prototype.resolvePrecedence = function (result, parser) {
           lLevel &&
           rLevel &&
           rLevel <= lLevel &&
-          !this.isRightAssociative(result.type)
+          (result.type !== result.right.type ||
+            !this.isRightAssociative(result.type))
         ) {
           // https://github.com/glayzzle/php-parser/issues/79
           // shift precedence
--- a/test/precedence.test.js
+++ b/test/precedence.test.js
@@ -89,6 +89,12 @@ describe("Test precedence", function () {
   it("test ?? right-associative", function () {
     shouldBeSame("1 ?? 2 ?? 3", "1 ?? (2 ?? 3)");
   });
+  it("test ?? precedence", function () {
+    shouldBeSame("1 ?? 2 and 3", "(1 ?? 2) and 3");
+  });
+  it("test ** precedence", function () {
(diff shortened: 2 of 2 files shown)


czosel:
Hi @ichiriac,
thanks for the review! `isRightAssociative` is only called from the one if branch that was changed. I also thought about changing the implementation of `isRightAssociative` first, but the function only does one thing: It takes an operator, and returns true if the operator is right-associtive. In order to tell if the "swap" should be performed in `resolvePrecedence`, the right-associativity should only be considered if the two operators have the same precedence level. This is equivalent to the two operators being different, because `**` and `??`, the only two right-associative operators PHP knows, have different precedence.

The bug only affected `**` and `??`, I added a test for the latter.

czosel:
@ichiriac friendly ping 🙃

ichiriac:
sorry, totally forgot this issue :smiley:

ichiriac:
The PR is merged, I'll try to fix #511 and release a patch, this week or next one

czosel:
No worries, thanks for merging! :tada:

PR #580, opened 2020-07-29
fix: parse nested shorthand lists as lists

### LINKED ISSUE #512: Nested shorthand lists parsed as array
Input:
```php
list($a, list($b, $c)) = array(1, array(2, 3));
[$a, [$b, $c]] = [1, [2, 3]];
```
The first line is parsed as `list` with a `list` inside, the second one as `list` with an `array` inside.

czosel:
... instead of arrays.

Fixes #512

### DIFF
--- a/src/parser/expr.js
+++ b/src/parser/expr.js
@@ -489,7 +489,7 @@ module.exports = {
       expr = this.read_scalar();
       if (expr.kind === "array" && expr.shortForm && this.token === "=") {
         // list assign
-        const list = this.node("list")(expr.items, true);
+        const list = this.convertToList(expr);
         if (expr.loc) list.loc = expr.loc;
         const right = this.next().read_expr();
         return result("assign", list, right, "=");
@@ -506,6 +506,27 @@ module.exports = {
 
     // returns variable | scalar
     return expr;
+  },
+
+  /**
+   * Recursively convert nested array to nested list.
+   */
+  convertToList: function (array) {
+    const convertedItems = array.items.map((entry) => {
+      if (
+        entry.value &&
+        entry.value.kind === "array" &&
+        entry.value.shortForm
+      ) {
+        entry.value = this.convertToList(entry.value);
+      }
+      return entry;
+    });
+    const node = this.node("list")(convertedItems, true);
+    if (array.loc) node.loc = array.loc;
+    if (array.leadingComments) node.leadingComments = array.leadingComments;
+    if (array.trailingComments) node.trailingComments = array.trailingComments;
+    return node;
   },
 
   /**
--- a/test/snapshot/__snapshots__/acid.test.js.snap
+++ b/test/snapshot/__snapshots__/acid.test.js.snap
@@ -7540,7 +7540,7 @@ next:
                     },
                   },
                   "unpack": false,
-                  "value": Array {
+                  "value": List {
                     "items": Array [
                       Entry {
                         "byRef": false,
@@ -7633,7 +7633,7 @@ next:
                         },
                       },
                     ],
-                    "kind": "array",
+                    "kind": "list",
                     "loc": Location {
                       "end": Position {
                         "column": 16,
--- a/test/snapshot/__snapshots__/list.test.js.snap
+++ b/test/snapshot/__snapshots__/list.test.js.snap
@@ -1020,6 +1020,358 @@ Program {
 }
 `;
 
+exports[`Test list expressions handles comments in nested shorthand lists 1`] = `
+Program {
+  "children": Array [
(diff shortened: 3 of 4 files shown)


test/snapshot/__snapshots__/acid.test.js.snap
```diff
@@ -7633,18 +7633,18 @@ next:
...
                     "loc": Location {
                       "end": Position {
-                        "column": 16,
+                        "column": 17,
                         "line": 141,
-                        "offset": 3034,
+                        "offset": 3035,
                       },
-                      "source": "[$b, ,$c]",
+                      "source": " ",
```
czosel, line 7643: Missing source - might be an issue?

test/snapshot/__snapshots__/list.test.js.snap
```diff
@@ -1630,18 +1745,18 @@ Program {
...
                 ],
-                "kind": "array",
+                "kind": "list",
                 "loc": Location {
                   "end": Position {
-                    "column": 19,
+                    "column": 24,
                     "line": 1,
-                    "offset": 19,
+                    "offset": 24,
```
czosel, line 1753: I'm not sure why the locations are different.

src/parser/expr.js
```diff
@@ -508,6 +508,23 @@ module.exports = {
...
+      if (
+        entry.value &&
+        entry.value.kind === "array" &&
+        entry.value.shortForm
+      ) {
+        entry.value = this.convertToList(entry.value.items);
+      }
+      return entry;
+    });
+    return this.node("list")(convertedItems, true);
```
ichiriac, line 525: @czosel this.node("...") => creates a node from the parsing context, should be used only when consuming tokens - as items were already processed, you need on created node to get in sync with locations & comments (available on array node)

```
convertToList: function (array) {
    const convertedItems = array.items.map((entry) => {
      if (
        entry.value &&
        entry.value.kind === "array" &&
        entry.value.shortForm
      ) {
        entry.value = this.convertToList(entry.value);
      }
      return entry;
    });
    const node = this.node("list")(convertedItems, true);
    node.loc = array.loc;
    node.leadingComments = array.leadingComments;
    node.trailingComments = array.trailingComments;
    return node;
```

test/snapshot/__snapshots__/acid.test.js.snap
ichiriac: yes, see next comment on locations

test/snapshot/__snapshots__/list.test.js.snap
ichiriac: see comment on locations

ichiriac (changes requested):
waiting to fix convertToList function

src/parser/expr.js
czosel: Hi @ichiriac, thanks for the detailed feedback! I made the change that fixes the `loc` issue, but I left out the comment part because I wasn't able to write a test that would show an issue with comments. Are you sure it is really needed? If yes, can you post an example code snipped that should show the issue? I tried `/* test */ [$a, [$b, $c]] = [1,[2,3]]` and `[$a, /* test */ [$b, $c]] = [1,[2,3]]`. Thanks again!

ichiriac: Hi @czosel, did you tries to enable comments flags by default the parser does not extracts comments. In principle without these statements comments eaten from node will not be available on created one.

Also try this one : `[$a, /* test */ [$b, $c]] =  [1,[2,3]] /* should not be eaten by newly created node ! */`

ichiriac: hi @czosel, still missing the comments statements, waiting for these before merging, other parts are clean

czosel: Hi @ichiriac, sorry about the delay - I hope I’ll find the time to finalize it this weekend!

ichiriac: no problemo dude, take your time, and tell me when you're ready to merge

czosel: Hi @ichiriac, I managed to take another look. I wasn't able to write a test for this because comments in our example either get assigned to `entry` or `assign` nodes, but never to `list`. I added the comment assignment as you suggested anyway - maybe there are situations where comments get attached to the list; and even if there currently is no such case, it possibly prevents future bugs if the comment handling ever changes :slightly_smiling_face: 

Let me know if you'd still change anything about this!

ichiriac (approved):
👍

src/parser/expr.js
ichiriac: thanks @czosel - I don't see also any usecase for now but your code is more solid this way as it covers better the api 😄

PR #671, opened 2021-02-23
update node version in CI

czosel:


czosel:
@ichiriac I'll go ahead and merge because it's a trivial change and it would be nice to get the CI to run again. Feel free to add feedback anyway!

ichiriac:
thanks @czosel I've added 12 & 13 versions also + cleaned & updated dependencies

PR #673, opened 2021-02-27
feat(php8): named arguments

### LINKED ISSUE #653: PHP8: Named arguments support
#486

czosel:
cc @Selion05

fixes #653

### DIFF
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -283,14 +283,27 @@ module.exports = {
   },
   /**
    * ```ebnf
-   *    argument_list ::= T_ELLIPSIS? expr
+   *    argument_list ::= T_STRING ':' expr | T_ELLIPSIS? expr
    * ```
    */
   read_argument: function () {
     if (this.token === this.tok.T_ELLIPSIS) {
       return this.node("variadic")(this.next().read_expr());
     }
-    return this.read_expr();
+    const name = this.text();
+    let res;
+    try {
+      res = this.read_expr();
+    } catch (e) {
+      // happens if named argument name equals keyword name, e.g.
+      // foo(array: $a) - `array` is not a valid expr
+      this.expect(":") && this.next();
+      return this.node("namedargument")(name, this.read_expr());
+    }
+    if (this.token === ":") {
+      return this.node("namedargument")(name, this.next().read_expr());
+    }
+    return res;
   },
   /**
    * read type hinting
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -290,20 +290,20 @@ module.exports = {
     if (this.token === this.tok.T_ELLIPSIS) {
       return this.node("variadic")(this.next().read_expr());
     }
-    const name = this.text();
-    let res;
-    try {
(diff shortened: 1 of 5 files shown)


czosel:
What I suggested in #673  didn't make sense because arguments are a list of `Expression` objects, where a `name` property doesn't make sense. This implements a new node kind `namedargument` instead.

src/parser/function.js
```diff
@@ -266,14 +266,27 @@ module.exports = {
...
    */
   read_argument: function () {
     if (this.token === this.tok.T_ELLIPSIS) {
       return this.node("variadic")(this.next().read_expr());
     }
-    return this.read_expr();
+    const name = this.text();
+    let res;
+    try {
+      res = this.read_expr();
```
ichiriac, line 279: hi, not sure that's the right way, you can also go ahead scan next tokens to see f it matches a T_STRING  + ':' and if not go for expr call

ichiriac (changes requested):
would be less error prone to use setState from lexer and check ahead with next
https://github.com/glayzzle/php-parser/blob/master/src/lexer.js#L339

czosel:
Hi @ichiriac, thanks for the review! Unfortunately I don’t really have time to work on this currently. If you prefer to make the changes yourself, go ahead 😉

czosel:
@ichiriac I tried to refactor this a you suggested, but it doesn't work when heredocs are used as function arguments:
```php
stringManipulator(<<<END
   a
  b
 c
END);
```
Can you point me in the right direction? :slightly_smiling_face:

czosel:
Alright, I found a solution / workaround: I only check ahead if the current token is T_STRING or a keyword - that way, the rollback is not needed for HEREDOC.

src/parser/function.js
czosel: fixed :slightly_smiling_face:

czosel:
@cseufert Can you take a look at these changes before we merge into the `php8` branch?

cseufert:
It appears that the current code does not follow the nikic parser, as it returns some arguments as thier immediate node, rather than returning an argument AST node (Arg in PHP-Parser). Is this something that we can break doing this update, or should we retain the backward compatibility? Also the function arguments return Entry for some of them, however this seems tied to an array entry, so I guess adding a name to this is not the best option.

As I understand it, backtracking is quite expensive in the parser, this patch will backtrack quite regularly, do we want to do a work around to avoid this? The issue here is the workaround will be quite complex.

czosel:
Hi @cseufert, thanks for the review! I think we should try to keep backwards compatibility as far as possible - in this case, the deviation from the nikic PHP parser must have already existed before PHP8.
Concerning performance: Interesting point, I don't really know about this. The original implementation was not very clean but didn't use backtracking, and @ichiriac asked me to rewrite it using backtracking. I'd be curious about his opinion on this though.

czosel:
I'll go ahead and merge this so that we can try it in the prettier plugin. @ichiriac as always, your feedback is still appreciated :slightly_smiling_face:

Issue #655, opened 2021-02-05
PHP8: Constructor property promotion support


Wulfheart:
#486

cseufert:
Please have a look at #685

genintho:
@czosel  I think this issue can be fixed. The readonly PR demonstrated that it is supported

czosel:
Totally, this landed in #697

PR #685, opened 2021-03-10
feat(php8): promoted properties

cseufert:
Basic impementation of promoted properties.

Added new AST node `promotedparameter` which is based on `parameter` with added `promote` property which can be `public`, `private`, or `protected`.

Is this the correct way to do this?

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -527,6 +527,7 @@ AST.prototype.checkNodes = function () {
   require("./ast/offsetlookup"),
   require("./ast/operation"),
   require("./ast/parameter"),
+  require("./ast/promotedparameter"),
   require("./ast/parentreference"),
   require("./ast/post"),
   require("./ast/pre"),
--- a/src/ast/promotedparameter.js
+++ b/src/ast/promotedparameter.js
@@ -9,8 +9,8 @@ const Declaration = require("./declaration");
 const KIND = "promotedparameter";
 
 /**
- * Defines a function parameter
- * @constructor Parameter
+ * Defines a promoted function parameter (for class constructor)
+ * @constructor PromotedParameter
  * @extends {Declaration}
  * @property {Identifier|null} type
  * @property {Node|null} value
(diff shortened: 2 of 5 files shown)


cseufert:
This should close #655

cseufert:
@czosel I think this approach should be changed in favour of how https://github.com/nikic/PHP-Parser/ handles it, and add a Param.flags that holds a int flag for visibility (public/private/protected) rather than a new ast node type. Ill add another PR for that when I get a chance

czosel:
For reference: PHP-Parser parses
```php
class __proto__ {
  public function constructor(public int $id, private $name, protected ServerRequestInterface $req) {}
}
```
as
```
array(                                                                                                                                                                                                                                                 
    0: Stmt_Class(                                                                                                                                                                                                                                     
        attrGroups: array(                                                                                                                                                                                                                             
        )                                                                                                                                                                                                                                              
        flags: 0                                                                                                                                                                                                                                       
        name: Identifier(                                                                                                                                                                                                                              
            name: __proto__                                                                                                                                                                                                                            
        )                                                                                                                                                                                                                                              
        extends: null                                                                                                                                                                                                                                  
        implements: array(                                                                                                                                                                                                                             
        )                                                                                                                                                                                                                                              
        stmts: array(                                                                                                                                                                                                                                  
            0: Stmt_ClassMethod(
                attrGroups: array(
                )
                flags: MODIFIER_PUBLIC (1)
                byRef: false
                name: Identifier(
                    name: constructor
                )
                params: array(
                    0: Param(
                        attrGroups: array(
                        )
                        flags: MODIFIER_PUBLIC (1)
                        type: Identifier(
                            name: int
                        )
                        byRef: false
                        variadic: false
                        var: Expr_Variable(
                            name: id
                        )
                        default: null
                    )
                    1: Param(
                        attrGroups: array(
                        )
                        flags: MODIFIER_PRIVATE (4)
                        type: null
                        byRef: false
                        variadic: false
                        var: Expr_Variable(
                            name: name
                        )
                        default: null
                    )
                    2: Param(
                        attrGroups: array(
                        )
                        flags: MODIFIER_PROTECTED (2)
                        type: Name(
                            parts: array(
                                0: ServerRequestInterface
                            )
                        )
                        byRef: false
                        variadic: false
                        var: Expr_Variable(
                            name: req
                        )
                        default: null
                    )
                )
                returnType: null
                stmts: array(
                )
            )
        )
```

I agree, a visibility flag is the best way to represent this :+1:

cseufert:
Use the implmentaton in #697

PR #686, opened 2021-03-10
feat(php8): match expression

cseufert:
Added new AST objects for Match and  MatchEntry


czosel:
PHP-Parser is parsing 
```php
match(trye) {
  0,1,2,3 => run(),
  default => null,
};
```
as
```
              expr: Expr_Match(
                cond: Expr_ConstFetch(
                    name: Name(
                        parts: array(
                            0: trye
                        )
                    )
                )
                arms: array(
                    0: MatchArm(
                        conds: array(
                            0: Scalar_LNumber(
                                value: 0
                            )
                            1: Scalar_LNumber(
                                value: 1
                            )
                            2: Scalar_LNumber(
                                value: 2
                            )
                            3: Scalar_LNumber(
                                value: 3
                            )
                        )
                        body: Expr_FuncCall(
                            name: Name(
                                parts: array(
                                    0: run
                                )
                            )
                            args: array(
                            )
                        )
                    )
                    1: MatchArm(
                        conds: null
                        body: Expr_ConstFetch(
                            name: Name(
                                parts: array(
                                    0: null
                                )
                            )
                        )
                    )
                )
            )
```
Notable differences:

- `arms` instead of `body` 
- `MatchArm` instead of `MatchEntry`
- `conds` instead of `keys`
- `default` is not parsed as a special keyword, but as `conds: null`

The PHP [documentation](https://www.php.net/manual/en/control-structures.match.php) also speaks of arms, it might be a good idea to follow this? Also, parsing `default` as `conds: null` seems like a simple solution to me - what do you think @cseufert ?

cseufert:
Thanks for the guidance on this, btw, what PHP-Parser project are you referencing?

I have updated the match statement, as per the code above.

czosel:
AST looks good! I'll try to take a look at the code later today. Here is the parser I was refering to: https://github.com/nikic/PHP-Parser

src/ast.js
```diff
@@ -486,6 +486,7 @@ AST.prototype.checkNodes = function () {
   require("./ast/declaration"),
   require("./ast/declare"),
   require("./ast/declaredirective"),
+  require("./ast/defaultkeyword"),
```
czosel, line 489: not needed anymore

src/ast/defaultkeyword.js
```diff
@@ -0,0 +1,23 @@
...
+ * @extends {Node}
+ */
+module.exports = Node.extends(KIND, function DefaultKeyword(
+  raw,
+  docs,
+  location
+) {
+  Node.apply(this, [KIND, docs, location]);
+  this.raw = raw;
+});
```
czosel, line 23: not needed anymore

czosel (commented):
Just some cleanup, otherwise looks good to me :+1:

src/ast/defaultkeyword.js
cseufert: Fixed

src/ast.js
cseufert: Fixed

cseufert:
@czosel I think this is now pretty much complete, however I havent checked for all syntax corner cases in tests, but the normal ones are in there.

czosel (approved):
:+1:

Issue #656, opened 2021-02-05
PHP8: Union types support


Wulfheart:
#486

czosel:
We're parsing `function (Foo $foo): Bar {}` as
```
{
      "kind": "function",
      "name": {
        "kind": "identifier",
        "name": false
      },
      "arguments": [
        {
          "kind": "parameter",
          "name": {
            "kind": "identifier",
            "name": "foo"
          },
          "value": null,
          "type": {
            "kind": "name",
            "resolution": "uqn",
            "name": "Foo"
          },
          "byref": false,
          "variadic": false,
          "nullable": false
        }
      ],
      "byref": false,
      "type": {
        "kind": "name",
        "resolution": "uqn",
        "name": "Bar"
      },
      "nullable": false,
      "body": {
        "kind": "block",
        "children": []
      }
```
PHP-Parser parses `function (Foo|Bar $foo): Baz|Buzz {};` as
```
expr: Expr_Closure(
            attrGroups: array(
            )
            static: false
            byRef: false
            params: array(
                0: Param(
                    attrGroups: array(
                    )
                    flags: 0
                    type: UnionType(
                        types: array(
                            0: Name(
                                parts: array(
                                    0: Foo
                                )
                            )
                            1: Name(
                                parts: array(
                                    0: Bar
                                )
                            )
                        )
                    )
                    byRef: false
                    variadic: false
                    var: Expr_Variable(
                        name: foo
                    )
                    default: null
                )
            )
            uses: array(
            )
            returnType: UnionType(
                types: array(
                    0: Name(
                        parts: array(
                            0: Baz
                        )
                    )
                    1: Name(
                        parts: array(
                            0: Buzz
                        )
                    )
                )
            )
            stmts: array(
            )
        )
```
We could introduce a `uniontype` node, e.g.
```
      "type": {
        "kind": "uniontype",
        "types": [{
          "kind": "name",
          "resolution": "uqn",
          "name": "Foo"
        }, {
          "kind": "name",
          "resolution": "uqn",
          "name": "Foo"
        }]
      },
```

See also https://github.com/php/php-src/blob/493524454d66adde84e00d249d607ecd540de99f/Zend/zend_language_parser.y#L650

cc @ichiriac

cseufert:
I have had a go at implementing this a changing type to an array, rather than a single value. Sorry I missed this before I started. I can update it to return a union type when there are multiple types.

alexander-akait:
Fixed in `php8` branch

PR #692, opened 2021-03-22
Support for PHP8 Union Types

cseufert:
This change makes all function & method parameters, return types, and class member variables parse types as a list (array), rather than single items.

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -309,7 +309,7 @@ module.exports = {
       nullable = true;
       this.next();
     }
-    let type = this.read_type();
+    let type = this.read_types();
     if (nullable && !type) {
       this.raiseError(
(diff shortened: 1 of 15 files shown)


cseufert:
This should close #656

cseufert:
I think #693 is more inline with what was proposed in ticket #656

cseufert:
I would not merge this and the v2 version

cseufert:
#693 is a cleaner implementation

czosel:
Oh sorry! I’ll revert this one later today.

czosel:
I just realized that GitHub shows this MR as merged because #693 actually contains the commits of this MR as well. The git history might not be very clean, but otherwise I think we're good :wink:

PR #693, opened 2021-03-22
feat(php8): union types with new AST node

cseufert:
This implementation should have much better bacward compatiblity than the other PR i submitted.


czosel:
Looks great! :100:

PR #697, opened 2021-03-29
feat(php8): promoted params v2

cseufert:
This patch is more inline with nikic PHP-Parser AST syntax, and should be 100% backward compatible with existing ast consumers.

### DIFF
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -208,6 +208,7 @@ module.exports = {
       this.next();
       nullable = true;
     }
+    const flags = this.read_promoted();
     types = this.read_types();
     if (nullable && !types) {
       this.raiseError(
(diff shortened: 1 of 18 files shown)


src/parser/function.js
```diff
@@ -244,6 +253,23 @@ module.exports = {
...
+      this.next();
+      return MODIFIER_PUBLIC;
+    } else if (this.token === this.tok.T_PROTECTED) {
+      this.next();
+      return MODIFIER_PROTECTED;
+    } else if (this.token === this.tok.T_PRIVATE) {
+      this.next();
+      return MODIFIER_PRIVATE;
+    } else {
+      0;
```
czosel, line 270: Should this be `return 0;`?

czosel:
Looks great otherwise :+1:

src/parser/function.js
cseufert: Thanks, i missed that

czosel:
Looks like you’ll still need to run prettier to make the linter happy..

czosel:
I went ahead and did the formatting fix myself in #702

PR #701, opened 2021-04-01
Fix named arguments comments

czosel:
I finally figured out what @ichiriac meant by looking ahead in https://github.com/glayzzle/php-parser/pull/673#discussion_r590342168 :relieved: 

cc @cseufert

### DIFF
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -294,14 +294,18 @@ module.exports = {
       this.token === this.tok.T_STRING ||
       Object.values(this.lexer.keywords).includes(this.token)
     ) {
-      const backup = [this.token, this.lexer.getState()];
-      const name = this.text();
-      this.next();
-      if (this.token === ":") {
-        return this.node("namedargument")(name, this.next().read_expr());
+      const lexerState = this.lexer.getState();
+      const nextToken = this.lexer.lex();
+      this.lexer.setState(lexerState);
+      if (nextToken === ":") {
+        if (this.version < 800) {
+          this.raiseError("PHP 8+ is required to use named arguments");
+        }
+        return this.node("namedargument")(
+          this.text(),
+          this.next().next().read_expr()
+        );
       }
-      this.lexer.tokens.push(backup);
-      this.next();
     }
     return this.read_expr();
   },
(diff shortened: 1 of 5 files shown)


ichiriac:
hi @czosel, I didn't had much time to check github messages, I'll be more active in few weeks. I'll take a look of your code and the last commit today.

src/parser/function.js
```diff
@@ -294,14 +294,15 @@ module.exports = {
       this.token === this.tok.T_STRING ||
       Object.values(this.lexer.keywords).includes(this.token)
     ) {
-      const backup = [this.token, this.lexer.getState()];
-      const name = this.text();
-      this.next();
-      if (this.token === ":") {
-        return this.node("namedargument")(name, this.next().read_expr());
+      const lexerState = this.lexer.getState();
+      const nextToken = this.lexer.lex();
```
ichiriac, line 298: 👍

test/snapshot/call.test.js
```diff
@@ -253,6 +253,15 @@ describe("Test call", function () {
     });
     expect(astErr).toMatchSnapshot();
   });
+  it("doesnt confused static methods with named arguments", function () {
+    const astErr = parser.parseEval(`foo(a::bar());`, {
+      parser: {
+        version: "8.0",
```
ichiriac, line 259: should be interesting to see behavior with `version: "7.2"`

ichiriac (commented):
perfect @czosel that's exactly how I would implement it

test/snapshot/call.test.js
czosel: Good point! I'll send a PR with a test.

czosel:
> hi @czosel, I didn't had much time to check github messages, I'll be more active in few weeks. I'll take a look of your code and the last commit today.

Hi @ichiriac, no worries at all!

When you do find the time again, it would we great if you could take a look at the other commits in the `php8` branch as well (@cseufert sent several PRs).

test/snapshot/call.test.js
czosel: done! I implemented the version check and added a test.

ichiriac:
@czosel I'll do it an release it under 3.1 when php8 will be fully implemented

PR #702, opened 2021-04-03
feat(php8): promoted params v2

czosel:
Copy of #697 with formatting fix applied.

### DIFF
--- a/src/ast/parameter.js
+++ b/src/ast/parameter.js
@@ -8,6 +8,13 @@
 const Declaration = require("./declaration");
 const KIND = "parameter";
 
+// eslint-disable-next-line no-unused-vars
+const MODIFIER_PUBLIC = 1;
+// eslint-disable-next-line no-unused-vars
+const MODIFIER_PROTECTED = 2;
+// eslint-disable-next-line no-unused-vars
+const MODIFIER_PRIVATE = 4;
+
 /**
  * Defines a function parameter
  * @constructor Parameter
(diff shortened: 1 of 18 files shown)


test/snapshot/class.test.js
```diff
@@ -129,15 +129,20 @@ describe("Test classes", function () {
...
-  it("Test class union properties", function () {
-    expect(
-      parser.parseEval(`
-      class Test {
-        static int|float $foo;
-        private ?Foo|Bar $bar;
-        public Repo|string|null $a;
-      }`)
-    ).toMatchSnapshot();
+  it("Test promoted class properties php 8", function () {
```
ichiriac, line 132: @czosel  seems that `it("Test class union properties", function () {` was removed, is there a reason ?

src/ast/parameter.js
```diff
@@ -8,6 +8,13 @@
 const Declaration = require("./declaration");
 const KIND = "parameter";
 
+// eslint-disable-next-line no-unused-vars
+const MODIFIER_PUBLIC = 1;
```
ichiriac, line 12: @czosel not sure thats a good approach, flags are better in memory usage but not readable, here I've used a different approch in other classes : https://github.com/glayzzle/php-parser/blob/master/src/ast/classconstant.js#L12

czosel:
@ichiriac I'll go ahead and cc @cseufert who implemented this 😉

PR #724, opened 2021-04-30
WIP php8-attributes

Selion05:
@ichiriac  @czosel 

The Current implementation can handle code like `<?php #[Attribute("aoeu",'stnh'),Attribute(),SUP] class Whatever {}`  as well as named properties `<?php #[Attribute(someProp: 'stnh')] class Whatever {}`. But it feels like i duplicated logic. Any feedback?

The tokens are equal to `token_get_all` in php

## todo
-  handle `\n`  in attributes 
- tests
-

### DIFF
--- a/src/lexer.js
+++ b/src/lexer.js
@@ -474,6 +474,7 @@ lexer.prototype.next = function () {
 
 // extends the lexer with states
 [
+  require("./lexer/attribute.js"),
   require("./lexer/comments.js"),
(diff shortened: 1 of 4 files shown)


cseufert:
I have extended this code to product attribute AST nodes for, still a WIP. See #745

cseufert:
In #745 I have added support for newlines, and arrays in parameters. Plus parsing into AST

PR #756, opened 2021-06-20
feat(PHP 8 danging comma): support dangling comma on function declaration

cseufert:
This should fix: https://github.com/prettier/plugin-php/issues/1759

### DIFF
--- a/test/snapshot/function.test.js
+++ b/test/snapshot/function.test.js
@@ -102,7 +102,6 @@ describe("Function tests", function () {
       {
         parser: {
           version: "8.0",
-          suppressErrors: true,
         },
       }
     );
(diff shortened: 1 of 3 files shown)


test/snapshot/function.test.js
```diff
@@ -91,4 +91,21 @@ describe("Function tests", function () {
...
+    const astErr = parser.parseEval(
+      `function (
+    $a,
+    $b,
+    $c,
+  ) {}`,
+      {
+        parser: {
+          version: "8.0",
+          suppressErrors: true,
```
czosel, line 105: Do we need to suppress errors here?

czosel (commented):
Looks great! Just one question

test/snapshot/function.test.js
cseufert: No we don't, I have removed that option.

czosel (approved):

PR #757, opened 2021-06-21
feat(eslint:curly) in PHP 8 branch enforce curly braces

cseufert:
Just an idea, to keep the code format consistent


czosel:
I'd like to hear @ichiriac's opinion about this - without his ok, I'd prefer not to introduce formatting changes between the `master` and `php8` branches.

czosel:
Landed in #863

PR #758, opened 2021-06-21
feat(eslint:curly multi-line) Enforce curly braces for multi-line block statements 

cseufert:
less heavy-handed eslint config alternative to #757


czosel:
Updated this to point to master in #863

PR #776, opened 2021-07-22
feat(php8 dangling comma): Support dangling comma in use-block

justim:
Allow for dangling commas to exist in the use-block of a closure.

This becomes valid:

```php
$foo = function() use ($bar,) {};
```

Walking over the variables is done in an external function and would need significant refactoring to allow for dangling commas, this might be too impactful.

This change is possible because `read_lexical_var` is only used for parsing the `use`-block.

### DIFF
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -150,8 +150,28 @@ module.exports = {
     return result;
   },
 
+  read_list_with_dangling_comma: function (item) {
+    const result = [];
+
+    while (this.token != this.EOF) {
+      result.push(item());
+      if (this.token == ",") {
+        this.next();
+        if (this.version >= 800 && this.token === ")") {
+          return result;
+        }
+      } else if (this.token == ")") {
+        break;
+      } else {
+        this.error([",", ")"]);
+        break;
+      }
+    }
+    return result;
+  },
+
   read_lexical_var_list: function () {
-    return this.read_list(this.read_lexical_var, ",");
+    return this.read_list_with_dangling_comma(this.read_lexical_var.bind(this));
   },
 
   /**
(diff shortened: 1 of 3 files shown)


src/parser/function.js
```diff
@@ -160,6 +160,12 @@ module.exports = {
    * ```
    */
   read_lexical_var: function () {
+    // a trailing separator is allowed in PHP8, which would put us right on the
+    // closing parenthesis. the ")" can be ignored here, the actually closing
+    // of the list is done in `read_lexical_vars`
+    if (this.version >= 800 && this.token === ")") {
+      return;
```
alexander-akait, line 167: Why return? I think it is wrong solution

justim: `read_list` moves past the dangling comma and runs this function. And the rest of this function expects a variable, but that will never come if there is a `)`, so we need to exit early when this is the case.

alexander-akait: I think we need put this in other place, before running this function, otherwise we have mixed logic

justim: I agree that this solution is not ideal.

The `read_list` function has no knowledge about the enclosing parenthesis, shall I create a specialized `read_list` for this purpose, like is done for the parameters of a function (`read_parameter_list`)?

Or do you have a different solution in mind?

test/snapshot/function.test.js
```diff
@@ -107,4 +107,13 @@ describe("Function tests", function () {
...
   });
+
+  it("test danging comma in closure use-block php 8.0", function () {
+    const astErr = parser.parseEval("$test = function () use ($one,) {}", {
+      parser: {
+        version: "8.0",
+      },
+    });
+    expect(astErr).toMatchSnapshot();
+  });
```
alexander-akait, line 118: Looks very better, let's add more tests here:
```
$example = function () use (&$message) {
    var_dump($message);
};
```
with/without trailing comma

alexander-akait (approved):
/cc @czosel

czosel (approved):
Looks good to me! Thanks @justim :100: 
cc @ichiriac

justim:
Thanks!

Issue #486, opened 2020-01-19
Add PHP 8 Support / WIP


ichiriac:
Prepare the parser to PHP 8 :

- [ ] https://wiki.php.net/rfc/union_types_v2

Work In Progress as RFC are not yet closed

Selion05:
https://wiki.php.net/rfc/nullsafe_operator

What is the status about php 8 support?

Wulfheart:
What’s new in PHP8? 

- Named arguments
- Attributes
- Constructor property promotion 
- Union types
- Match expression
- Nullsafe operator

I’d suggest creating an separate issue for each if these features. 

Compiled from https://www.php.net/releases/8.0/en.php

cseufert:
There are now patches available for:
- named arguments #673 
- property promotion #685 
- union types #656 
- match expression #686
- nullsafe operator #662 (merged?)

so really the work just needs to focus on attributes, however I am not sure the right way to update the lexer to support Attributes.

czosel:
Hi @cseufert 
Awesome work! #662 has been merged in a temporary `php8` branch because I don't know the codebase well enough to just merge stuff into master before @ichiriac took a look as well. If you like, you can point your PRs to the `php8` branch as well, and I'll try to find some time for a first review :)

cseufert:
I have changed the following PR's to merge to the php8 branch:
- #693 Union Types
- #686 Match expression
- #685 Promoted properties

I am going to have a go at getting this stuff working in prettier at some point, and am looking foward to having someone who know the codebase have a review of the code. I feel it could benefit a lot from being ported to typescript, but I guess thats a totally different problem.

czosel:
Awesome! As I said before, reviews in the parser are also quite difficult for me - I'm more comfortable on the prettier side. Either way I'll try to do my best to help, it would be awesome to have full php8 support in the prettier plugin :-)

Eldair:
Hello,

There is no tracking for non-capturing catches - `} catch (Exception) {` `Parse Error : syntax error, unexpected ')'`

MasonD:
Is this the correct place to remark on an error I noticed with parsing of promoted constructor params? php-parser complains about a syntax error when a promoted constructor param is marked as nullable with a question mark
```
public function __construct(public ?string $nullable) { }
```

cseufert:
[...]

PR #785, opened 2021-08-16
Fixed parsing of optional promoted properties when type is nullable

cseufert:
There was a logic error, I was trying to pass the promotion keyword after the nullable type token. I have added a test, and re-ordered the parsing correctly.

This should fix:
https://github.com/glayzzle/php-parser/issues/486#issuecomment-897346495


czosel (approved):
LGTM 👍

PR #788, opened 2021-09-06
PHP8 Attrributes repeating

cseufert:
Reported in prettier:
https://github.com/prettier/plugin-php/pull/1761#issuecomment-913770153

> Hello i have found a bug in this version when creating a class with multiple methods and you add an attribute just in the first one... all the following methods will have the same attribute... like the last one is stuck for all the rests...

I have added a test case, and a fix for it.

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -105,6 +105,7 @@ module.exports = {
       if (this.token === this.tok.T_FUNCTION) {
         // reads a function
         result.push(this.read_function(false, flags, attrs));
+        attrs = [];
       } else if (
         this.token === this.tok.T_VARIABLE ||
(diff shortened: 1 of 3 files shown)


cseufert:
@czosel are you able to merge this, and then release a new prettier?

fernando-rivas-smtp (approved):

alexander-akait (approved):

czosel (approved):

czosel:
@ichiriac do you think you could take a look at the php8 branch? It would be great if we could merge it back into master and get it released before the two branches diverge too much.

@cseufert until then, we’ll have to stick to explicit commit references..

PR #789, opened 2021-09-08
Add enum support

MaartenStaa:
I came into contact with this project after adopting `@pretter/plugin-php`. Considering future proof-ness, I figured adding enum support would be a good step forward.

Note that I dove in head-first, so please let me know if my approach at any point does not make sense. In particular, what's the difference between T_CLASS and T_CLASS_C? I mirrored this for the enums, but perhaps that does not make sense. Similarly, I'm not sure the places I've added `T_ENUM`/`T_ENUM_C` in `Parser.entries` make sense. Some comments there for future contributors on what this is for and what each map in the entry means might be useful.

Looking forward to your feedback!


alexander-akait (approved):

src/parser/enum.js
```diff
@@ -0,0 +1,126 @@
...
+ * @authors https://github.com/glayzzle/php-parser/graphs/contributors
+ * @url http://glayzzle.com
+ */
+"use strict";
+
+module.exports = {
+  /*
+   * reading a class
+   * ```ebnf
+   * class ::= class_scope? T_CLASS T_STRING (T_EXTENDS NAMESPACE_NAME)? (T_IMPLEMENTS (NAMESPACE_NAME ',')* NAMESPACE_NAME)? '{' CLASS_BODY '}'
```
cseufert, line 12: Looks to be direct copy of class docs, same on 44-46

src/parser/enum.js
```diff
@@ -0,0 +1,126 @@
...
+      if (this.token === this.tok.T_DOC_COMMENT) {
+        result.push(this.read_doc_comment());
+        continue;
+      }
+
+      // check T_USE trait
+      if (this.token === this.tok.T_USE) {
+        result = result.concat(this.read_trait_use_statement());
+        continue;
+      }
```
cseufert, line 67: I dont think you can have `use` statements in traits.

cseufert: Actually never mind, learned something new today.

src/parser/enum.js
MaartenStaa: Adjusted. Not 100% sure if the one on line 46 is correct now though.

cseufert (approved):
Looks great, LGTM

czosel:
@MaartenStaa Nothing specific - all good then!

@ichiriac maybe we should mark the codeclimate stage as optional?

ichiriac:
yes @czosel, the code climate is optional

czosel:
@MaartenStaa sorry it took so long to get this merged - i took your PR, changed it to point to `master` again now that `php8` has been merged, applied some formatting changes, and merged it: #862

PR #792, opened 2021-10-05
Fix method flags in src location

cseufert:
Fix method flags in src location parsing classes and interfaces

This will fix problems when using prettier-ignore in front of a function and that static/private/etc/ disappears.

Not sure if this is the best way to do this, but it appears to work correctly.

Can @czosel merge this into php8 branch?

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -128,21 +128,6 @@ const AST = function (withPositions, withSource) {
   this.withSource = withSource;
 };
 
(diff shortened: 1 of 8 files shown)


cseufert:
@ichiriac or @czosel are we able to get this merged, so we can get ahttps://github.com/prettier/plugin-php/issues/1648 fixed?

PR #813, opened 2021-10-29
Typescript fixes

ssigwart:
- Fix "always"
- Allow null bodies on loops (e.g. `while (doSomething());`)
- Fix boolean typo


ichiriac (approved):
:+1:

PR #865, opened 2022-02-05
Add basic support for readonly properties

genintho:
PHP 8.1 added support for readonly properties. 
https://wiki.php.net/rfc/readonly_properties_v2

```php
class Test {
    public readonly string $prop;
 
    public function __construct(string $prop) {
        $this->prop = $prop;
    }
}
```

```php
class Test {
 
    public function __construct(public readonly string $prop) {
    }
}
```


genintho:
I would love some feedback to know if this is a good approach.

src/parser/class.js
```diff
@@ -172,19 +172,25 @@ module.exports = {
        */
       function read_variable_declaration() {
         const result = this.node("property");
+        let readonly = false;
+        if (this.text() === "readonly") {
```
cseufert, line 176: You should be able to check for the token T_READ_ONLY here rather than text.

cseufert (commented):
You should be able to add the same check just after the promoted properties are found in the parameter parser see: src/parser/function.js:239

Also the other snapshots should be updated since you have updated the Parameter Node

genintho:
@cseufert I followed your suggestion. IMO this is overloading a bit much the concept of "Parameters" and creating a "class constructor" and "class constructor parameter" would be cleaner. But that's way more work :D

cseufert:
@genintho Yes, we have been mostly following how this parser does it: https://github.com/nikic/PHP-Parser

Generally, the nodes and attributes are following the same names. I guess what it really needed is some sort of parse error if we encounter promoted or read-only param and we are not in a constructor, however that's probably outside of the scope of these changes.

genintho:
@cseufert I added something, let me know what you think of the approach

src/parser/class.js
```diff
@@ -122,6 +122,7 @@ module.exports = {
       } else if (
         allow_variables &&
         (this.token === this.tok.T_VARIABLE ||
+          this.token === this.tok.T_READ_ONLY ||
```
czosel, line 125: I think we should add a version check here, such that parsing read only properties only works if `version >= 810`

test/snapshot/__snapshots__/graceful.test.js.snap
```diff
@@ -837,18 +838,19 @@ Program {
           "kind": "propertystatement",
           "properties": Array [
             Property {
-              "attrGroups": Array [],
+              "attrGroups": null,
```
czosel, line 841: This diff and the one right below looks a bit off, doesn't it?

src/parser/function.js
```diff
@@ -228,15 +230,28 @@ module.exports = {
...
     let parameterName = null;
     let value = null;
     let types = null;
     let nullable = false;
+    let readonly = false;
     let attrs = [];
     if (this.token === this.tok.T_ATTRIBUTE) attrs = this.read_attr_list();
     const flags = this.read_promoted();
+
+    if (this.token === this.tok.T_READ_ONLY) {
```
czosel, line 244: This line needs a version check as well

test/snapshot/__snapshots__/graceful.test.js.snap
genintho: The test in question is

```javascript
 it("test class", function () {
      expect(test.parseEval("class foo { foo const A = 1 ")).toMatchSnapshot();
    });
```

My understanding of this test suite is to make sure that the parser does not throw an error when providing invalid PHP and when setting the option `suppressErrors` to true. 

If that understanding is correct, I am not sure I need to try to "fix" how an invalid PHP class gets parsed.

czosel: Ah, I didn't check what the test was actually doing. Sounds reasonable!

czosel (approved):
:100:

czosel:
Published in `v3.1.0-beta.2` :tada:

genintho:
Cool Thanks!
Do you know why all the update of this package are published behind a "beta" tag as opposed to a regular version number?

czosel:
Thank you for implementing this! :wink: 

We're currently doing beta releases because quite many changes have landed since the last stable release. If no significant issues are reported in the coming weeks, we'll do a regular release.

PR #871, opened 2022-02-15
Fix several small typing issues

cpiber:
Fixes several places where the types were wrong, and makes the `Literal` sub-types narrow the `value`.

Note: These were all extracted from the AST where I thought to check because the given typing does not make sense, I did not check the code.

I also noticed that there are changes that were made several months ago but still aren't published, when can we expect an update?

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -77,6 +77,7 @@ declare module "php-parser" {
      * Defines a boolean value (true/false)
      */
     class Boolean extends Literal {
+        value: boolean;
     }
     /**
      * A break statement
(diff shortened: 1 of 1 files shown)


czosel:
The beta releases are on npm as well :slightly_smiling_face: 
See https://www.npmjs.com/package/php-parser/v/3.1.0-beta.2

czosel (approved):
Looks good! :+1:

czosel:
Released as `v3.1.0-beta.5` :tada:

PR #880, opened 2022-02-20
Default version is PHP 8.1

genintho:
Lexer and Parser should be updated to the same version.

In addition, PHP 8.1 translate to `801` . See [comment](https://github.com/glayzzle/php-parser/blob/main/src/index.js#L49) and [implementation of version parsing](https://github.com/glayzzle/php-parser/blob/main/src/index.js#L84).
IMO it is very confusing and error prone.


czosel:
> In addition, PHP 8.1 translate to 801

Oh, that's confusing :see_no_evil: Good catch (both of them)!

czosel:
Released in `v3.1.0-beta.4` :tada:

PR #883, opened 2022-02-21
[8.1] - Add support for intersection type

genintho:
Copied the approach for union type support.

But one thing was tricky: parameter pass by reference. 

```php
function a(Foo&Bar &$something)
```

The  loop finding all the type in the intersection/union return a null type for the last `&`. By doing so, the "active" token in the parser is now on `$something`, and so[ the call to `is_reference()`](https://github.com/glayzzle/php-parser/blob/main/src/parser/function.js#L265) is returning false. In order to fix this, I added the concept of a `previous_token` and check its value in is_reference. 

I am not a big fan of this solution because it make a strict assumption on what is the order of the call of the functions. I tried to [add a new token type](https://github.com/glayzzle/php-parser/blob/main/src/lexer/tokens.js#L254) for the union and intersection type, but could not make it work. So I thought this was good enough with the unit test coverage 🤷

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -522,6 +522,7 @@ AST.prototype.checkNodes = function () {
   require("./ast/include"),
   require("./ast/inline"),
   require("./ast/interface"),
+  require("./ast/intersectiontype"),
   require("./ast/isset"),
   require("./ast/label"),
   require("./ast/list"),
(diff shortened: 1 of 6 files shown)


genintho:
@czosel An other addition 😄

genintho:
@czosel a friendly ping, whenever you have time, no rush.

czosel:
@genintho sorry about the delay - I hope I’ll find some time tomorrow.

cseufert:
@czosel and @genintho I make and alternative version without tracking the previous token. See #893

PR #888, opened 2022-02-27
chore: try cachegrind for reproducibility

czosel:


czosel:
Looks promising, merging for further testing

PR #893, opened 2022-02-27
Parse Intersection types (alt version)

cseufert:
I have changed the implementation to not require the previous token


genintho (approved):
I think this version is better than the one I proposed: it does not require to keep an additional state.

czosel (approved):
Awesome! :100:

czosel:
Released as `v3.1.0-beta.5` :tada:

PR #917, opened 2022-04-29
Fix intersection types with by-ref variable

justim:
Previously the following code would produce the wrong AST:

```php
function test(foo|bar &$input) {}
```

It would convert the union to an intersection. (See the [Prettier playground])

This PR changes this behavior by checking for the next token first, before setting the "mode" of the multiple types.

PHP does not allow mixing `|` and `&`, an additional change here is that the parser now follows that rule.

[Prettier playground]: https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDaRTtrMMEEJvBAAHgB+YzAAHUwtqUFMGIxMAAIdcYAKKXR0AB8ONggjgDIAEmI5mABKI+B2kHagA

### DIFF
--- a/src/parser/function.js
+++ b/src/parser/function.js
@@ -287,29 +287,55 @@ module.exports = {
     return result;
   },
   read_types() {
+    const MODE_UNSET = "unset";
+    const MODE_UNION = "union";
+    const MODE_INTERSECTION = "intersection";
+
     const types = [];
-    let isIntersection = false;
+    let mode = MODE_UNSET;
     const type = this.read_type();
     if (!type) return null;
+
+    // we have matched a single type
     types.push(type);
+
+    // is the current token a:
+    // - | for union type
+    // - & for intersection type (> php 8.1)
     while (this.token === "|" || (this.version >= 801 && this.token === "&")) {
-      if (this.token === "&") {
-        isIntersection = true;
-        const nextToken = this.peek();
+      const nextToken = this.peek();
+
+      if (
+        nextToken === this.tok.T_ELLIPSIS ||
+        nextToken === this.tok.T_VARIABLE
+      ) {
+        // the next token is part of the variable (or the variable itself),
+        // we're not gonna match anymore types
+        break;
+      }
+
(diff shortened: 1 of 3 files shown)


src/parser/function.js
```diff
@@ -287,29 +287,55 @@ module.exports = {
...
+    let mode = MODE_UNSET;
     const type = this.read_type();
     if (!type) return null;
+
+    // we have matched a single type
     types.push(type);
+
+    // is the current token a:
+    // - | for union type
+    // - & for interaction type (> php 8.1)
```
czosel, line 304:
```suggestion
    // - & for intersection type (> php 8.1)
```

src/parser/function.js
```diff
@@ -287,29 +287,55 @@ module.exports = {
...
+        // the next token is part of the variable (or the variable itself),
+        // we're not gonna match anymore types
+        break;
+      }
+
+      if (mode === MODE_UNSET) {
+        // are we in union or interaction "mode"
+        mode = this.token === "|" ? MODE_UNION : MODE_INTERSECTION;
+      } else {
+        // it is not possible to mixed "modes"
```
czosel, line 321:
```suggestion
        // it is not possible to mix "modes"
```

czosel (commented):
Looks great, thanks! :100: Just found two minor typos in comments.

justim:
Thanks, I fixed the typos. 👍

czosel:
Released in `3.1.0-beta.6` :tada:

PR #937, opened 2022-05-25
Add PHP 8.1 first class callable support

### LINKED ISSUE #936: PHP 8.1 - First class callable support
I am using prettier PHP plugin in latest version which uses the beta7 version releases yesterday and most things work fine now but first class callable support is still missing.

Examples:
```php
array_map(trim(...), $names);
```

```php
array_map($this->foo(...), $values);
```

MaartenStaa:
Fixes #936

### DIFF
--- a/test/snapshot/__snapshots__/function.test.js.snap
+++ b/test/snapshot/__snapshots__/function.test.js.snap
@@ -189,6 +189,128 @@ Program {
 }
 `;
 
+exports[`Function tests first class callable support 1`] = `
+Program {
+  "children": Array [
+    ExpressionStatement {
+      "expression": Assign {
+        "kind": "assign",
+        "left": Variable {
+          "curly": false,
+          "kind": "variable",
+          "name": "callable",
+        },
+        "operator": "=",
+        "right": Call {
+          "arguments": Array [
+            229,
+          ],
+          "kind": "call",
+          "what": Name {
+            "kind": "name",
+            "name": "strlen",
+            "resolution": "uqn",
+          },
+        },
+      },
+      "kind": "expressionstatement",
+    },
+    ExpressionStatement {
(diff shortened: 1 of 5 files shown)


test/snapshot/__snapshots__/function.test.js.snap
```diff
@@ -189,6 +189,128 @@ Program {
...
+        "kind": "assign",
+        "left": Variable {
+          "curly": false,
+          "kind": "variable",
+          "name": "callable",
+        },
+        "operator": "=",
+        "right": Call {
+          "arguments": Array [
+            229,
```
czosel, line 229: this represents the `...` (ellipsis), right? The token codes are usually not part of our output - do you know how the official PHP parser represents this?

czosel (commented):
Looks great already! Just one question.

test/snapshot/__snapshots__/function.test.js.snap
MaartenStaa: Correct, that's the value of `T_ELLIPSIS`, fair point about the token codes not usually being used in the AST output. I did check how nikic/php-parser represents this, and it looks like this:

```
array(
    0: Stmt_Expression(
        expr: Expr_Assign(
            var: Expr_Variable(
                name: x
            )
            expr: Expr_FuncCall(
                name: Name(
                    parts: array(
                        0: test
                    )
                )
                args: array(
                    0: VariadicPlaceholder(
                    )
                )
            )
        )
    )
)
```

So then maybe we want a `VariadicPlaceholder` AST node. What do you think?

czosel: Sounds good! :+1:

czosel (approved):
Looks great! 👍

PR #940, opened 2022-05-26
fix: enum as name

### LINKED ISSUE #896: `Enum` is treated as a reserved word, even before PHP 8.1
Input: `class Enum {}`
Output: syntax error, unexpected 'Enum' (T_ENUM), expecting T_STRING

The official PHP parser seems to be able to handle this example even after PHP 8.1 - if that turns out to be too difficult to fix here, maybe should consider adding the keyword only for PHP >= 8.1?

Also see https://github.com/glayzzle/php-parser/pull/895 for a failing test.

czosel:
This is a first naive attempt to fix #896. There are probably several cases that I missed, but I'd like to get some feedback - is this something that can be solved in the lexer instead?

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -21,7 +21,7 @@ module.exports = {
       this.next();
       return null;
     }
-    this.next().expect(this.tok.T_STRING);
+    this.next().expect([this.tok.T_STRING, this.tok.T_ENUM]);
     let propName = this.node("identifier");
     const name = this.text();
     this.next();
@@ -381,7 +381,7 @@ module.exports = {
       this.next();
       return null;
     }
-    this.next().expect(this.tok.T_STRING);
+    this.next().expect([this.tok.T_STRING, this.tok.T_ENUM]);
     let propName = this.node("identifier");
     const name = this.text();
     this.next();
@@ -465,7 +465,7 @@ module.exports = {
       this.next();
       return null;
     }
-    this.next().expect(this.tok.T_STRING);
+    this.next().expect([this.tok.T_STRING, this.tok.T_ENUM]);
     let propName = this.node("identifier");
     const name = this.text();
     this.next();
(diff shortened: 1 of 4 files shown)


MaartenStaa:
@czosel I think this is something we should try to solve in the lexer instead. As an example, given the following PHP code:

```php
enum foo {}
class enum {}
```

this is the output from `token_get_all`:

```
T_ENUM#336 ('enum')
T_WHITESPACE#392 (' ')
T_STRING#262 ('foo')
T_WHITESPACE#392 (' ')
string(1) "{"
string(1) "}"
T_WHITESPACE#392 ('
')
T_CLASS#333 ('class')
T_WHITESPACE#392 (' ')
T_STRING#262 ('enum')
T_WHITESPACE#392 (' ')
string(1) "{"
string(1) "}"
```

In other words, whether the string "enum" is interpreted as `T_STRING` or `T_ENUM` seems to be contextual. It looks like the way PHP handles this is by looking ahead to the next tokens:

```lex
/*
 * The enum keyword must be followed by whitespace and another identifier.
 * This avoids the BC break of using enum in classes, namespaces, functions and constants.
 */
<ST_IN_SCRIPTING>"enum"{WHITESPACE}("extends"|"implements") {
	yyless(4);
	RETURN_TOKEN_WITH_STR(T_STRING, 0);
}
<ST_IN_SCRIPTING>"enum"{WHITESPACE}[a-zA-Z_\x80-\xff] {
	yyless(4);
	RETURN_TOKEN_WITH_IDENT(T_ENUM);
}
```

czosel:
@MaartenStaa I had a feeling this would be the way to go, but didn't check how PHP itself handles this - thanks! :+1: I'm not sure if I'll find the time to implement this in the coming days, so if you (or anyone else) wants to take over that's totally fine by me :wink:

czosel:
Superseded by #941

PR #941, opened 2022-05-28
fix: enum as identifier name

### LINKED ISSUE #896: `Enum` is treated as a reserved word, even before PHP 8.1

czosel:
This seems much cleaner :-)

Fixes #896

### DIFF
--- a/src/lexer/tokens.js
+++ b/src/lexer/tokens.js
@@ -32,6 +32,18 @@ module.exports = {
       }
     }
 
+    if (id === this.tok.T_ENUM) {
+      const initial = this.offset;
+      let ch = this.input();
+      while (ch == " ") {
+        ch = this.input();
+      }
+
+      const is_enum = this.is_LABEL_START();
+      this.unput(this.offset - initial);
+      return is_enum ? this.tok.T_ENUM : this.tok.T_STRING;
+    }
+
     if (this.offset < this.size && id !== this.tok.T_YIELD_FROM) {
       // If immediately followed by a backslash, this is a T_NAME_RELATIVE or T_NAME_QUALIFIED.
       let ch = this.input();
--- a/src/lexer/tokens.js
+++ b/src/lexer/tokens.js
@@ -33,6 +33,9 @@ module.exports = {
     }
 
     if (id === this.tok.T_ENUM) {
+      if (this.version < 801) {
+        return this.tok.T_STRING;
+      }
       const initial = this.offset;
       let ch = this.input();
       while (ch == " ") {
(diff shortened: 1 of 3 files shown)


src/lexer/tokens.js
```diff
@@ -32,6 +32,21 @@ module.exports = {
...
+        return this.tok.T_STRING;
+      }
+      const initial = this.offset;
+      let ch = this.input();
+      while (ch == " ") {
+        ch = this.input();
+      }
+
+      const is_enum = this.is_LABEL_START();
+      this.unput(this.offset - initial);
```
MaartenStaa, line 46: I agree, this is much nicer! Thinking of how the PHP lexer handles this, I think this misses an edge case though. I think you need to look ahead through the entire label to check if it's not equal to `implements` or `extends`, to handle:

```php
class Enum extends Foo {}
class Enum implements Bar {}
```

MaartenStaa (changes requested):

src/lexer/tokens.js
czosel: Good call, I totally missed that! Fixed.

src/lexer/tokens.js
```diff
@@ -41,10 +42,16 @@ module.exports = {
       while (ch == " ") {
         ch = this.input();
       }
+      let isEnum = false;
+      if (this.is_LABEL_START()) {
+        while (this.is_LABEL()) {
+          ch += this.input();
+        }
+        isEnum = !ch.startsWith("extends") && !ch.startsWith("implements");
```
MaartenStaa, line 50: Why `startsWith` rather than `!==`? I think you need to check for equality to cover e.g. `enum extendsType`. And come to think of it, I think you also need `ch.toLowerCase()`, since this is perfectly valid:

```php
class Enum exTEnds Foo {}
```

MaartenStaa (changes requested):

src/lexer/tokens.js
czosel: I was using startswith because `ch` contains one character too much (the first non-label character is appended in the last iteration. But you're right of course, `startsWith` is not what we need. Fixed, and also added `toLowerCase`.

MaartenStaa: Aah, gotcha. Yeah, that makes sense. LGTM now!

czosel: 🎉

czosel:
Released in `v3.1.0-beta.9` :tada:

PR #960, opened 2022-06-24
Handle property types using T_NAME_* tokens.

### LINKED ISSUE #957: Error when type hinting class properties with fully qualified classname
```php
class TestClass {
    public \DateTime $property;
}
```

results in the error `Parse Error : syntax error, unexpected '\DateTime' (T_NAME_FULLY_QUALIFIED)`

MaartenStaa:
Fixes #957


czosel (approved):
:+1:

PR #978, opened 2022-07-31
fix #968

### LINKED ISSUE #968: Error in types.d.ts TypeScript definitions
php-parser version 3.1.0-beta.11
tsc version 4.7.4

```
node_modules/php-parser/types.d.ts:327:9 - error TS2416: Property 'value' in type 'Encapsed' is not assignable to the same property in base type 'Literal'.
  Type 'EncapsedPart[]' is not assignable to type 'string | number | boolean | Node | null'.

327         value: EncapsedPart[];
            ~~~~~


Found 1 error in node_modules/php-parser/types.d.ts:327
```

Possible solution: extend Encapsed class directly from Expression and add `raw: string;` directly to Encapsed class.

ytetsuro:


ytetsuro:
Sorry.
I will fix JSDoc and come back out.

PR #1046, opened 2022-12-07
feat: support for readonly class

genintho:
Is this approach ok?

According to https://php.watch/versions/8.2/readonly-classes

> Abstract classes final classes can also be declared readonly. The order of the keywords does not make a difference.

I know some tests will fail, I am showing this to verify that my approach is ok.


czosel:
Hi @genintho - the general approach looks good to me! :+1:

genintho:
Not sure why CI is breaking now. Running the typescript command locally show no difference. 🤔 I must be missing something.

genintho:
@czosel this is ready for review.

I had to install `@types/node` to fix CI. In the CI, we globally install the most recent version of typescript, which conflict with the version implicitly installed in the repo.

I would love an alternative solution, but I already waste a couple of hours on this.

czosel:
@genintho it seems the typescript issue was due to outdated transitive dependencies (see [here](https://github.com/microsoft/TypeScript/issues/51567) for details) - I just fixed it in #1050 by regenerating the lock file. Can you revert the changes to package.json and yarn.lock and rebase on `main`?

genintho:
@czosel happy new year! 😄 

I did what you asked. We should be good to go.

czosel:
@genintho thank you, all the best for you as well! :tada: 

LGTM :+1:

czosel:
Released in `v3.1.3` :tada:

PR #1116, opened 2023-06-07
chore(CI): bump node version

czosel:
Node 14 is EOL. @cseufert strictly speaking this is a breaking change - should we treat it as such?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,47 +9,47 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x, 20.x]
 
     steps:
-    - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


cseufert:
We are not yet breaking on node14 are we? Do we need to hard drop support for it, or should we wait until a dependency/feature can no longer support node 14?

czosel:
Totally, we aren’t. Shall we leave node 14 in CI until it breaks then?

cseufert:
Yes, I think we leave it, and once it breaks our next release will drop support for it.

.github/workflows/test.yml
```diff
@@ -9,47 +9,47 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
```
cseufert, line 12:
```suggestion
        node-version: [14.x, 16.x, 18.x, 20.x]
```

Keep node 14.x support for now

cseufert (changes requested):

cseufert (approved):
LGTM

Issue #1130, opened 2024-03-08
Heredoc fails to parse escaped newlines


cseufert:
Source of this problem: https://github.com/prettier/plugin-php/issues/2015#issuecomment-1982167227

Unable to parse snippet like this:
```
<?php
if (true) {
    echo <<<STR
    \na
    STR;
}
```

PR #1131, opened 2024-03-08
Fix problem with heredoc parsing \n as newlines, not as literal string

cseufert:
I am trying to fox #1130 however need some help with this one, for the prettier plugin we want to keep the original intent, and not parse the \n as a newline, so we can rebuild the original string in the output. 

Do I create a special AST token for \n in a HEREDOC, or should ignore the \n as a special case, and parse it as a regular string?


czosel:
Hmm, I don't have any intuition on this. Might be worth checking how this is handled in https://github.com/nikic/PHP-Parser?

PR #1132, opened 2024-05-07
Update usegroup.js

medhamza911:
Correcting name attributes in definition documents.


cseufert:
Looks like you may also need to update  `types.d.ts`

PR #1138, opened 2024-08-14
Handle the ::class static keyword as a special case in parser/variable.js

mreiden:
Handle T_Class as a special case when it comes after T_DOUBLE_COLON in recursive_variable_chain_scan

This solves https://github.com/glayzzle/php-parser/issues/1135
```
$arrayVariable[0]::class;
$classVariable->classProp::class
```

### DIFF
--- a/src/parser/variable.js
+++ b/src/parser/variable.js
@@ -216,8 +216,15 @@ module.exports = {
             this.error();
           }
 
+          // Handle special "::class" static keyword
+          let classKeyword = null;
+          if (this.peek() === this.tok.T_CLASS) {
+              this.next();
+              classKeyword = this.node("identifier")(this.text());
+              this.next();
+          }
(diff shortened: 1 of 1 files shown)


ichiriac (changes requested):
The issue comes from the way PHP understands `identifier` https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L325

From a certain version, identifiers are composed also by reserved and semi reserved keywords :
https://github.com/glayzzle/php-parser/blob/main/src/parser/variable.js#L161

Here an example on how it's handled from php 7.0 :
https://github.com/glayzzle/php-parser/blob/80eb9460679188f76166dd25a9a2666b8ff129aa/src/parser/class.js#L584

jorgsowa:
Alternative way of solution with tests: https://github.com/glayzzle/php-parser/pull/1172

PR #1139, opened 2024-11-06
chore: update dependencies

czosel:


czosel:
CI is currently failing with errors like this:
```
Error: node_modules/@types/eslint-scope/index.d.ts(50,5): error TS2416: Property 'writeExpr' in type 'Reference' is not assignable to the same property in base type 'Reference'.
```
I have no typescript experience - any pointers are appreciated!

cseufert:
I am pretty sure we do not want `@types/eslint-scope` to be installed, as it seems to be out of date, and eslint seems to have the types included already

cseufert:
ok on further investigation, it appears that webpack imports @types/eslint-scope however the actualy eslint package looks to have updated typescript types. I am not sure how to solve this, without a PR to webpack to update its use of the eslint-scope types to consume directly from eslint.

cseufert:
Well rolling back webpack to 5.96.0 to avoid this fix: https://github.com/webpack/webpack/issues/18916

has resulted in 
```
Error: node_modules/@types/node/module.d.ts(108,13): error TS2386: Overload signatures must all be optional or required.

```

which make no sense either.

czosel:
@cseufert Thanks for looking into this! I tried following up on the remaining error, and it seems that installing `@types/node` fixes the issue: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/70562#discussioncomment-10666769

czosel:
@cseufert CI seems to pass now. Let me know if this is ready to be merged from your side :-)

cseufert (approved):
LGTM

Issue #1133, opened 2024-05-09
Fails to parse classes with typed constants


davidrans:
It looks like files that have typed constants are not parseable by the latest version (3.1.5).

Input file:
```
<?php

class Test {
   private const int MY_CONST = 3;
}
```
Error:
```
SyntaxError: Parse Error : syntax error, unexpected 'MY_CONST' (T_STRING), expecting '=' on line 4
    at Parser.raiseError (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser.js:349:17)
    at Parser.error (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser.js:397:15)
    at Parser.expect (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser.js:577:10)
    at Parser.read_constant_declaration (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/class.js:261:18)
    at Parser.read_list (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/utils.js:76:33)
    at Parser.read_constant_list (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/class.js:237:24)
    at Parser.read_class_body (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/class.js:117:32)
    at Parser.read_class_declaration_statement (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/class.js:32:30)
    at Parser.read_top_statement (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/statement.js:52:21)
    at Parser.read_start (/home/user/Code/node_modules/.pnpm/php-parser@3.1.5/node_modules/php-parser/src/parser/main.js:18:19) {
  lineNumber: 4,
  fileName: 'eval',
  columnNumber: 21
}
```

Kenneth-Sills:
[...]

PR #1140, opened 2024-11-25
Fix Typed Class Constants parsing

genintho:
Fix https://github.com/glayzzle/php-parser/issues/1133

- make sure all variation of class constant are still being supported
- Throw a clear error when the parser detect typed class constants by the PHP version configured does not match

### DIFF
--- a/test/snapshot/__snapshots__/attributes.test.js.snap
+++ b/test/snapshot/__snapshots__/attributes.test.js.snap
@@ -308,18 +308,7 @@ Program {
       "attrGroups": [],
       "body": [
         ClassConstant {
-          "attrGroups": [
-            AttrGroup {
-              "attrs": [
-                Attribute {
-                  "args": [],
-                  "kind": "attribute",
-                  "name": "B",
-                },
-              ],
-              "kind": "attrgroup",
-            },
-          ],
+          "attrGroups": undefined,
           "constants": [
             Constant {
               "kind": "constant",
@@ -335,7 +324,18 @@ Program {
           ],
           "final": false,
           "kind": "classconstant",
-          "nullable": false,
+          "nullable": [
+            AttrGroup {
+              "attrs": [
+                Attribute {
+                  "args": [],
+                  "kind": "attribute",
+                  "name": "B",
+                },
+              ],
+              "kind": "attrgroup",
+            },
+          ],
           "type": null,
           "visibility": "",
         },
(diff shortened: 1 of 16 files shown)


test/snapshot/__snapshots__/attributes.test.js.snap
```diff
@@ -335,7 +324,18 @@ Program {
...
+            AttrGroup {
+              "attrs": [
+                Attribute {
+                  "args": [],
+                  "kind": "attribute",
+                  "name": "B",
+                },
+              ],
+              "kind": "attrgroup",
+            },
```
cseufert, line 337: This does not make sense either, nullable should be a bool value? looks like attrGroups and nullable are switched around somewhere

cseufert (changes requested):
It looks like the refactor has significantly changed the AST output around attrgroups, as the attrGroups should not be affected, also I would expect nullable to be a boolean type.

genintho:
@cseufert Yeah there are bugs in the logic.
Can you look at this different PR I made that fix some other bugs I found while working on this https://github.com/glayzzle/php-parser/pull/1141

genintho:
I figured out the core issue: the AST ClassConstant object `constants` which is an array of AST "Constant" object.

```
ClassConstant {
          "attrGroups": undefined,
          "constants": [
            Constant {
              "kind": "constant",
              "name": Identifier {
                "kind": "identifier",
                "name": "CON_2",
              },
              "value": String {
                "isDoubleQuote": true,
                "kind": "string",
                "raw": ""Hello world!"",
                "unicode": false,
                "value": "Hello world!",
              },
            },
          ],
          "final": false,
          "kind": "classconstant",
          "nullable": [],
          "type": null,
          "visibility": "",
        },
```

The code is reading a list of constant, to support this syntax.

```php
class Foo {
    const PRI1 = 'baz', PRI2 = 'baz2';
}
```

But the implementations made the assumption that all constant have the same type/nullable: type and nullable was added to "ClassConstant", not "Constant". 

That's incorect, because this code is valid:

```
class Foo {
	private const PRI1 = 'baz', string PRI2 = 'baz2', int PRI3 = 5;
}
```

The last commit I pushed, 94d406975c251192c653dc8284c2cd7a31195c6e keep the type on the "ClassConstant" object and create only 1 constant object per ClassConstant. That works, but it is a less accurate representation of the code being parsed. 

Now I am going to fix that by moving the type to the AST "Constant" object.

test/snapshot/__snapshots__/constantstatement.test.js.snap
```diff
@@ -53,6 +57,8 @@ Program {
             "kind": "identifier",
             "name": "CONSTANT",
           },
+          "nullable": null,
+          "type": undefined,
```
genintho, line 61: If we do not want those to be part of the AST, we need to completly decouple the ClassConstant object from the Constant object.

genintho:
@cseufert code is ready for another review

genintho:
Actually, they are still bug. I will reopen when it is fully ready.

PR #1141, opened 2024-11-25
Consistent returns for `read_variable_declaration`

genintho:
While working on adding support for property hooks, I realized that the function parsing class property does not always call the "return" function with the same parameters: in some instance,  the parameter `readonly` is not being passed done. 

This PR fix that by having 1 call to return despite multiple branches in the logic.

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -204,13 +204,10 @@ module.exports = {
 
         let value = null;
 
-        if (this.token === ";" || this.token === ",") {
-          // no-op
-        } else if (this.token === "=") {
+        this.expect([",", ";", "="]);
+        if (this.token === "=") {
           // https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L815
           value = this.next().read_expr();
-        } else {
-          this.expect([",", ";"]);
         }
         return result(propName, value, readonly, nullable, type, attrs || []);
       },
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -210,7 +210,7 @@ module.exports = {
           // https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L815
           value = this.next().read_expr();
         } else {
-          this.expect([",", ";", "="]);
+          this.expect([",", ";"]);
         }
         return result(propName, value, readonly, nullable, type, attrs || []);
       },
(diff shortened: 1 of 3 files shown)


src/parser/class.js
```diff
@@ -201,22 +201,18 @@ module.exports = {
...
-            this.next().read_expr(),
-            readonly,
-            nullable,
-            type,
-            attrs || [],
-          );
+          value = this.next().read_expr();
         } else {
           this.expect([",", ";", "="]);
-          return result(propName, null, nullable, type, attrs || []);
```
genintho, line 218: here, readonly is not passed.

test/snapshot/__snapshots__/graceful.test.js.snap
```diff
@@ -300,19 +300,19 @@ Program {
...
           "properties": [
             Property {
-              "attrGroups": null,
+              "attrGroups": [],
               "kind": "property",
               "name": Identifier {
                 "kind": "identifier",
                 "name": "onst",
               },
-              "nullable": Name {
```
genintho, line 309: The snapshot show that what was set was incorrect: instead of a boolean value we had some objects.

src/parser/class.js
```diff
@@ -201,22 +201,18 @@ module.exports = {
...
-            propName,
-            this.next().read_expr(),
-            readonly,
-            nullable,
-            type,
-            attrs || [],
-          );
+          value = this.next().read_expr();
         } else {
           this.expect([",", ";", "="]);
```
cseufert, line 213: I think we need to drop the `"="` in th eexpect, we already handled that with the previous if condition.
```suggestion
          this.expect([",", ";"]);
```

cseufert (changes requested):
Looks good, if we could just streamline the if else when looking for a default value, they we can merge this.

src/parser/class.js
genintho: "," and ";" are handled by the first if.
I think this line is here to show an error message indicating that we expect one of those 3 things. 

I figured this out while working on adding support for property hooks.

cseufert (approved):
LGTM

genintho:
I do not have permission to merge 😄 .

czosel (approved):

Issue #1142, opened 2024-11-25
8.4: property hooks


genintho:
- [ ] get - simple
```php
    public string $email {
       get => 'mailto:' . $this->email;
    }
```

- [ ] get - block
```php
       get {
           'mailto:' . $this->email;
        }
```

- [ ] set - full block form
```php
set ($value) { $this->[propertyName = $value }
```

- [ ] block form with implicit $value
```php
set { $this->[propertyName = $value }
```

- [ ] expression form with explicit $value
```php
set ($value) => {expression}
```

- [ ] expression form with implicit $value
```php
set => {expression}// expression form with implicit $value
```

- [ ] get by reference
```php
&get => { }
```

- [ ] default value
```php
public string $role = 'anonymous' {
    set {
        Roles::from($value);
        $this->role = $value;
    }
}
```

- [ ] final
```php
class StandardUser
{
    public string $email {
        final set {
           if (! filter_var($value, FILTER_VALIDATE_EMAIL, FILTER_FLAG_EMAIL_UNICODE)) {
               throw new InvalidArgumentException('Invalid email');
           }
           $this->email = $value;
        }
    }
}
```

- [ ] abstract
```
abstract class StandardUser
{
    abstract public string $email { get; }
}
```

- [ ] interface - get
```
interface I
{
    // An implementing class MUST have a publicly-readable property,
    // but whether or not it's publicly settable is unrestricted.
    public string $readable { get; }
}
```

- [ ] interface - set
```
interface I
{ 
    // An implementing class MUST have a publicly-writeable property,
    // but whether or not it's publicly readable is unrestricted.
    public string $writeable { set; }
}
```

- [ ] interface - get + set
```
interface I
{
    // An implementing class MUST have a property that is both publicly
    // readable and publicly writeable.
    public string $both { get; set; }
}
```

PR #1136, opened 2024-08-06
feat: add support for PHP 8.3 typed class constants

Kenneth-Sills:
[Related RFC](https://wiki.php.net/rfc/typed_class_constants).

Closes #1133

PR #1143, opened 2024-12-01
[WIP] - 8.4 Property hooks

genintho:
Fixes https://github.com/glayzzle/php-parser/issues/1142

### DIFF
--- a/src/ast/property.js
+++ b/src/ast/property.js
@@ -18,6 +18,7 @@ const KIND = "property";
  * @property {boolean} readonly
  * @property {boolean} nullable
  * @property {Identifier|Array<Identifier>|null} type
+ * @propert {PropertyHooks[]|null} hooks
  * @property {AttrGroup[]} attrGroups
  */
 module.exports = Statement.extends(
@@ -28,6 +29,7 @@ module.exports = Statement.extends(
     readonly,
     nullable,
     type,
+    hooks,
     attrGroups,
     docs,
     location,
@@ -38,6 +40,7 @@ module.exports = Statement.extends(
     this.readonly = readonly;
     this.nullable = nullable;
     this.type = type;
+    this.hooks = hooks;
     this.attrGroups = attrGroups;
   },
 );
--- a/src/ast/property.js
+++ b/src/ast/property.js
@@ -18,7 +18,7 @@ const KIND = "property";
  * @property {boolean} readonly
  * @property {boolean} nullable
  * @property {Identifier|Array<Identifier>|null} type
- * @propert {PropertyHooks[]|null} hooks
+ * @propert {PropertyHook[]|null} hooks
  * @property {AttrGroup[]} attrGroups
  */
 module.exports = Statement.extends(
--- a/test/snapshot/__snapshots__/class.test.js.snap
+++ b/test/snapshot/__snapshots__/class.test.js.snap
@@ -1200,19 +1200,19 @@ Program {
           "kind": "propertystatement",
           "properties": [
             Property {
-              "attrGroups": null,
+              "attrGroups": [],
               "kind": "property",
               "name": Identifier {
                 "kind": "identifier",
                 "name": "id",
               },
-              "nullable": TypeReference {
+              "nullable": false,
+              "readonly": true,
+              "type": TypeReference {
                 "kind": "typereference",
                 "name": "int",
                 "raw": "int",
               },
-              "readonly": false,
-              "type": [],
               "value": null,
             },
           ],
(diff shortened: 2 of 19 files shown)


genintho:
@cseufert I would love your feedback on this PR.

test/snapshot/class.test.js
```diff
@@ -182,19 +182,19 @@ describe("Test classes", function () {
...
+    expect(() => {
+      parser.parseEval(
+        `
       class Bob {
         public function foo(public readonly int $id) {}
       }`,
-      {
-        parser: {
-          version: "8.1",
-          suppressErrors: true,
```
cseufert, line 193: Should this be changed?

types.d.ts
```diff
@@ -777,6 +777,16 @@ declare module "php-parser" {
     type: Identifier | Identifier[] | null;
     attrGroups: AttrGroup[];
   }
+  /**
+   * Defines a class property hook getter & setter
+   */
+  class PropertyHook extends Statement {
```
cseufert, line 783: I am not sure Statement is the best type to extends, probably better with a Function type.

This parser directly etends the Node type
https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/PropertyHook.php

test/snapshot/__snapshots__/class.test.js.snap
```diff
@@ -1158,163 +1173,7 @@ Program {
...
-      "kind": "error",
-      "line": 4,
-      "message": "Parse Error : syntax error, unexpected '}' on line 4",
-      "token": "'}'",
-    },
-  ],
-  "kind": "program",
-}
-`;
+exports[`Test classes Test that readonly method parameters are throwing errors 1`] = `"readonly properties can be used only on class constructor on line 3"`;
```
cseufert, line 1176: This looks like a regression.

src/ast/property.js
```diff
@@ -28,6 +29,7 @@ module.exports = Statement.extends(
     readonly,
     nullable,
     type,
+    hooks,
```
cseufert, line 32: It would probably make sense to put hooks last as this would be less breaking of a change for existing code. Also making the default [] would make the type signature simpler, and reduce ambiguity between an empty list and null, as an empty property list is not valid syntax.

cseufert (changes requested):
This looks very close, thanks for getting this working.

I noticed a couple of things, the major one being making the hooks an empty list rather than null, and having the constructor param last, rather than in-between existing params.

test/snapshot/__snapshots__/class.test.js.snap
genintho: This is related to the change you commented in the class.test file https://github.com/glayzzle/php-parser/pull/1143/files#r1865229710

Instead of letting the parser getting lost and generate gibberish errors, the exception is fired and serialized, using `toThrowErrorMatchingSnapshot`.

src/ast/property.js
genintho: Adding the `hooks` parameter at the end of the list would be inconsistent with how things have been done so far.

The PR introducing support for typed class constant added type "in the middle". https://github.com/glayzzle/php-parser/pull/1136/files#diff-33d9ef64a624f7abad7378a94559475184322ac87b7c864326beea672657adbcR34

genintho: @cseufert I do not think it is possible to put hooks as the last parameters: when we [create a node](https://github.com/glayzzle/php-parser/pull/1143/files#diff-fc169768643555846d9df28c56bdfeebd845afe68e2d045596b56ad7b915a9acR221-R229), the parameter `docs` and `location` are passed in by some other layer.

genintho: @czosel any opinion on this?

test/snapshot/class.test.js
cseufert: The array of errors thing is to allow parsing for tools like prettier where they can still format some of the code, even though the complete file is not valid AST (in theory at least, not sure that it works in practice)

src/ast/property.js
cseufert: Yer not last last, every node has docs and location last, this is automatically populated.

However if 3rd party code is replacing/creating new AST nodes, then this will cause some undefined behavior.

test/snapshot/__snapshots__/class.test.js.snap
cseufert: @czosel Whats your opinion on the change in the error handling?

src/ast/property.js
czosel: It's true that we didn't look out for this in #1136, but I guess if there is no other argument against adding it at last position (before docs and location) it would still make sense to do so - same goes for the default `[]`.

PR #1147, opened 2024-12-04
Fix Typed Class Constants parsing 

genintho:
Class using untyped constant are breaking the parser when using 8.3

```
class Foo { 
const CONSTANT = "Hello world!"; 
}
```

The problem came from `read_optional_type` which was returning a "type" even if the statement did not have a type.


czosel (approved):
LGTM, thanks for fixing this @genintho!

Issue #1144, opened 2024-12-02
8.4 - Asymmetric Visibility


genintho:
RFC - https://wiki.php.net/rfc/asymmetric-visibility-v2

```php
class Foo
{
    public private(set) string $bar = 'baz';
}
```

PR #1148, opened 2024-12-04
[WIP] - 8.4 Asymmetric  Visibility support

genintho:
https://github.com/glayzzle/php-parser/issues/1144

### DIFF
--- a/src/ast/classconstant.js
+++ b/src/ast/classconstant.js
@@ -8,11 +8,6 @@
 const ConstantStatement = require("./constantstatement");
 const KIND = "classconstant";
 
-const IS_UNDEFINED = "";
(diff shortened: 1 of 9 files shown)


genintho:
@czosel @cseufert This is work in progress, but I would like to hear your thoughts on the approach.

I tried to create a new token for `protected(set)` and `private(set)` but gave up after a bit as I could not find a good spot to do this.

I ended creating an intermediate object to store the various flags as it made things simpler to understand and removes a lot of duplication.

One thing I keep going back and forth is if the AST we create should match 100% the code or if it should match the intent of the code. For example `public private(set)` is the same as `private(set)`. So do we want the AST to be same in both situation or not?

Issue #1151, opened 2025-01-12
Syntax Error on multilined class constant with comment


shufo:
/cc @genintho

## Context
- php-parser https://github.com/glayzzle/php-parser/releases/tag/v3.2.2

## Problem
- class constant Syntax error introduced in 3.2.2 (#1147)
  - 3.2.1 does not occur this syntax error
  - https://github.com/prettier/plugin-php/blob/5f75fea7979af4dd530ea8cac827fb6314264621/tests/comments/classconstant.php#L16-L24
- It occured when class constant is multilined and comment is inserted before `=`. 

e.g.

```php
<?php

$var = new class {
    public
    const
    CONSTANT
    // Comment 3
    =
    'string';
};

```

### How to reproduce

- create test file

`vim test.js`

```js
const engine = require(".");

// initialize a new parser instance
const parser = new engine({
  // some options :
  parser: {
    extractDoc: true,
    phpVersion: "8.3",
  },
  ast: {
    withPositions: true,
    withSource: true,
  },
});

const target = `
<?php

$var = new class {
    public
    const
    CONSTANT
    // Comment 3
    =
    'string';
};
`;

// Retrieve the AST from the specified source
parser.parseCode(target);
```

- execute

```bash
$ node test.js
/home/shuhei/develop/php-parser/src/parser.js:357
    throw err;
    ^

SyntaxError: Parse Error : syntax error, unexpected '=' on line 11
    at Parser.raiseError (/home/shuhei/develop/php-parser/src/parser.js:349:17)
    at Parser.error (/home/shuhei/develop/php-parser/src/parser.js:397:15)
    at Parser.expect (/home/shuhei/develop/php-parser/src/parser.js:577:10)
    at Parser.read_constant_declaration (/home/shuhei/develop/php-parser/src/parser/class.js:262:16)
    at Parser.read_list (/home/shuhei/develop/php-parser/src/parser/utils.js:76:33)
    at Parser.read_constant_list (/home/shuhei/develop/php-parser/src/parser/class.js:234:24)
    at Parser.read_class_body (/home/shuhei/develop/php-parser/src/parser/class.js:117:32)
    at Parser.read_new_expr (/home/shuhei/develop/php-parser/src/parser/expr.js:769:28)
    at Parser.read_expr_item (/home/shuhei/develop/php-parser/src/parser/expr.js:353:21)
    at Parser.read_expr (/home/shuhei/develop/php-parser/src/parser/expr.js:18:19) {
[...]

PR #1152, opened 2025-01-14
fix: 🐛 syntax error on multiline class constant with comment

shufo:
## Context
- fixes: #1151 

## Problem
- see: #1151 

## Solution
Skip comments and whitespace when peeking as peek() is just peeking next token without consider comment. I don't know this is proper solution but it works as expected in my environment.

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -358,7 +358,7 @@ module.exports = {
       this.next();
     }
 
-    if (this.peek() === "=") {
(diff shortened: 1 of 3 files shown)


czosel:
@shufo sorry about the late response! We're looking to fix the CI in #1157.

czosel:
Released in `v3.2.4` :tada:

shufo:
@czosel 
Thank you!

PR #1156, opened 2025-06-20
fix: valid_after_heredoc_73 missing the concatenation operator

### LINKED ISSUE #1155: Syntax error when end of heredoc does not have a space between it and the concatenation operator
This is a short one for reproduction.

This one throws (`SyntaxError: Parse Error : syntax error, expecting T_END_HEREDOC on line 4`)
```php
<?php
echo <<<EOS

EOS. "";
```

This does not.
```php
<?php
echo <<<EOS

EOS . "";
```

Both are valid PHP code, with the same net effect. Though, the examples are basically as minimal as can be, to the point it would be nonsensical in real code. Replace `""` with a variable and add some meat to heredoc, then you get something reasonable. Like something you might get in a test, like where I found this.

By the looks of it, it's something added in PHP 7.3, and was missed in #506, so should be easy to add to https://github.com/glayzzle/php-parser/blame/c702357f7435af363051a947b64d1e064cca014b/src/lexer/strings.js#L10 and have a test added.

Spice-King:
Fixes #1155, sprinkled in some tests too relating to flexible nowdocs as well. One line fix, 130 lines of tests/fixtures.

### DIFF
--- a/src/lexer/strings.js
+++ b/src/lexer/strings.js
@@ -16,6 +16,7 @@ const valid_after_heredoc_73 = valid_after_heredoc.concat([
   "/",
   "=",
   "!",
+  ".",
 ]);
 
 module.exports = {
--- a/test/snapshot/__snapshots__/nowdoc.test.js.snap
+++ b/test/snapshot/__snapshots__/nowdoc.test.js.snap
@@ -27,6 +27,104 @@ c",
 }
 `;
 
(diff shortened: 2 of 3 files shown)


czosel (approved):
LGTM!

czosel:
Released in `v3.2.4` :tada:

Issue #1160, opened 2025-07-09
New without parenthesis followed by array lookup does not parse


czosel:
This should be valid in PHP 8.4:

```php
$var = new foo()[0];
```
But currently triggers an error:
```
    SyntaxError: Parse Error : syntax error, unexpected '[', expecting ';' on line 1

      347 |   message += " on line " + this.lexer.yylloc.first_line;
      348 |   if (!this.suppressErrors) {
    > 349 |     const err = new SyntaxError(
          |                 ^
      350 |       message,
      351 |       this.filename,
      352 |       this.lexer.yylloc.first_line,

      at Parser.raiseError (src/parser.js:349:17)
      at Parser.raiseError [as error] (src/parser.js:397:15)
      at Parser.error [as expectEndOfStatement] (src/parser.js:508:10)
      at Parser.expectEndOfStatement [as read_statement] (src/parser/statement.js:418:14)
      at Parser.read_statement [as read_top_statement] (src/parser/statement.js:78:21)
      at Parser.read_top_statement [as read_start] (src/parser/main.js:18:19)
      at Parser.read_start [as parse] (src/parser.js:301:22)
      at Engine.parse [as parseEval] (src/index.js:146:22)
      at Object.parseEval (test/snapshot/class.test.js:266:24)
```
cc @genintho

czosel:
Fixed in `v3.2.5` 🎉

PR #1162, opened 2025-07-10
fix: parse new without parens as `call` instead of `bin`

### LINKED ISSUE #1161: new without parenthesis is parsed as `bin`
While it should be parsed as `call`. See https://github.com/prettier/plugin-php/pull/2422#issuecomment-3052829472 for details.

cc @genintho

czosel:
Fixes #1161 and #1160 

cc @claytonrcarter

### DIFF
--- a/src/parser/class.js
+++ b/src/parser/class.js
@@ -386,7 +386,7 @@ module.exports = {
     return [nullable, type];
   },
 
-  peekSkipComments: function () {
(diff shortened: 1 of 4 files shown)


cseufert (approved):
Looks good.

czosel:
Fixed in `v3.2.5` 🎉

PR #1164, opened 2025-07-18
Initial support for PHP 8.5 Pipe Operator

cseufert:
I think this should work, not sure how smart we should be about rejecting non-callables in right side of pipeline operator.


czosel (approved):
@cseufert LGTM, sorry this was lying around for so long! I think it's ok to keep this simple (no elaborate callable-checking).

PR #1168, opened 2025-11-13
Add support for anonymous readonly classes

olivereisenhut:
This adds support for anonymous classes that are readonly.
e.g `new readonly {}`.

### DIFF
--- a/src/parser/expr.js
+++ b/src/parser/expr.js
@@ -764,10 +764,15 @@ module.exports = {
     }
     const attrs = this.read_attr_list();
     const isReadonly = this.token === this.tok.T_READ_ONLY;
-    if (
-      this.token === this.tok.T_CLASS ||
-      (isReadonly && this.next().token === this.tok.T_CLASS)
-    ) {
+    if (isReadonly) {
+      if (this.version < 803) {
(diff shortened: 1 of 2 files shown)


src/parser/expr.js
```diff
@@ -763,7 +763,11 @@ module.exports = {
       return result(newExp, args);
     }
     const attrs = this.read_attr_list();
-    if (this.token === this.tok.T_CLASS) {
+    const isReadonly = this.token === this.tok.T_READ_ONLY;
+    if (
+      this.token === this.tok.T_CLASS ||
+      (isReadonly && this.next().token === this.tok.T_CLASS)
+    ) {
```
czosel: Let's add a version check, and a slight refactor of the control flow along the lines of this:
```suggestion
    const isReadonly = this.token === this.tok.T_READ_ONLY;
    if (isReadonly) {
      if (this.version < 830) {
        // throw exception
      }
      this.next();
    }
    if (this.token === this.tok.T_CLASS) {
```

czosel (commented):
Thanks for contributing @olivereisenhut! This looks great, just a tiny tweak and we can get this merged.

PR #1232, opened 2026-03-25
fix: enforce underscore separator restrictions in hex, binary, and octal literals

jorgsowa:
This PR fixes the lexer to reject invalid underscore separator usage in prefixed numeric literals (0x…, 0b…, 0o…), matching PHP's actual behavior. Previously, the lexer silently accepted things like 0x1__2 (consecutive underscores) or 0x1_ (trailing underscore) as valid tokens. The PR adds tests for 6 invalid patterns, all of which now correctly produce parse errors instead of being silently accepted.

The four separate consume methods (consume_HNUM, consume_BNUM, consume_ONUM, consume_LNUM) were consolidated:
  - consume_prefixed_digits(isValid) — new shared method that handles the underscore-tracking logic (rejects consecutive __ and trailing _) for all prefixed literals
  - consume_hexadecimal() — thin wrapper calling consume_prefixed_digits(this.is_HEX)
  - consume_binary() — thin wrapper calling consume_prefixed_digits with an inline 0/1/_ check
  - consume_octal() — thin wrapper calling consume_prefixed_digits(this.is_OCTAL)
  - consume_decimal_digits() — rename of consume_LNUM

### DIFF
--- a/src/lexer/numbers.js
+++ b/src/lexer/numbers.js
@@ -23,7 +23,7 @@ module.exports = {
       if (ch === "x" || ch === "X") {
         ch = this.input();
         if (ch !== "_" && this.is_HEX()) {
-          return this.consume_HNUM();
+          return this.consume_hexadecimal();
         } else {
           this.unput(ch ? 2 : 1);
         }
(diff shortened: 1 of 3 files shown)


src/lexer/numbers.js
```diff
@@ -123,49 +123,44 @@ module.exports = {
       return this.tok.T_DNUMBER;
     }
   },
-  // read hexa
-  consume_HNUM() {
+  consume_prefixed_digits(isValid) {
```
czosel, line 126: Nitpick: I'd assume a boolean when I see `isValid`, shall we go with something along the lines of `validationFunc` or `isValidCheck`?

czosel:
Apart from the nitpick, LGTM :+1:

PR #1236, opened 2026-03-30
fix: correct unary minus/plus/tilde precedence

jorgsowa:
Unary -, +, and ~ shared the same precedence key as binary - and + (level 15), placing them below * (level 16). This caused -20 * 5 + 10 to be parsed as -(20 * 5 + 10) instead of (-20) * 5 + 10.

Add u-, u+, u~ keys to the precedence table at the ! level (17) and look them up in resolvePrecedence so unary operators correctly bind tighter than multiplicative operators.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -149,7 +149,7 @@ AST.precedence = {};
   ["<<", ">>"],
   ["+", "-", "."],
   ["*", "/", "%"],
-  ["!"],
+  ["!", "u-", "u+", "u~"], // u- etc. are unary variants; higher than * so -20*5 parses as (-20)*5
   ["instanceof"],
   ["cast", "silent"],
   ["**"],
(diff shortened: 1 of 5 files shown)


src/ast.js
```diff
@@ -149,7 +149,7 @@ AST.precedence = {};
   ["<<", ">>"],
   ["+", "-", "."],
   ["*", "/", "%"],
-  ["!"],
+  ["!", "u-", "u+", "u~"], // u- etc. are unary variants; higher than * so -20*5 parses as (-20)*5
```
czosel: Not sure how relevant in practice, but according to https://www.php.net/manual/en/language.operators.precedence.php the unary operators have even higher precedence than `instanceof`. Would it make sense to reflect this here?
Also, tangentially, are we lacking `++` and `--`?

jorgsowa: Good point. I changed the precedence of instanceof. Pre and post increment operators, we can tackle in another branch.

czosel: Hmm, could it be that you moved it too far? `instanceof` has lower precedence than the unary uperators, but higher precedence than the arithmetic operators (+, -, ., *, /, %).

jorgsowa: That's right. Operator `!` should be split away from unary arithmetic operators.

czosel (approved):
LGTM now 👍

czosel:
Released in `v3.7.0` :tada:

PR #1271, opened 2026-05-26
fix: default read visibility to public for shorthand asymmetric syntax

jorgsowa:
Fix shorthand asymmetric visibility syntax to correctly default read visibility to public (e.g., private(set) now means public property with private setter).


alexander-akait (approved):

czosel:
Released in `v3.7.0` :tada:

PR #1272, opened 2026-05-26
fix: clone operator precedence

jorgsowa:
The clone operator was incorrectly consuming binary operators. For example:
`clone $obj + 1`  was parsed as  `clone ($obj + 1)` instead of the correct: `(clone $obj) + 1`


alexander-akait (approved):

czosel:
Released in `v3.7.0` :tada:

